PR #43, opened 2017-04-14
Tatari clear all

ben-burlingham:
DELETE needed.


filters/tatari/src/TatariApi.js
```diff
@@ -9,3 +9,6 @@ export const get = (url, params) =>
 
 export const patch = (url, params, config) =>
   axios.patch(url, params, {...CSRFHEADER, ...config });
+
+export const deleteHTTP = (url, params, config) =>
```
chrisarevir: for consistency, can we change this to destroy?

PR #52, opened 2017-04-26
small style fixes for tatari filters

nickwmurphy:
Matching of styles with Rapu


filters/tatari/src/Tatari.scss
```diff
@@ -243,6 +235,7 @@ $searchH: 26px;
 .activeCheckbox {
   cursor: pointer;
   margin-right: 10px;
+  flex-shrink: 0;
```
chrisarevir: Please alphabetize 😢

PR #94, opened 2017-05-24
Ripanga ongroupcheck update key

brianyan:
tableData currently follows the schema below. Adjusted group select to match the naming convention. 
screen shot 2017-05-24 at 2 37 28 pm


chrisarevir:
If there's anything in the sandbox that could be updated, could ya please make sure it is up to date?

Also, by updating it in this way, it seems like we could actually expose the desired "key" to use in this situation. so a valueField prop that tells that function which key to use. Just my 2 cents.

brianyan:
The sandbox reflects the current naming convention so it technically is updated (there were errors popping up when selecting onGroupCheck before the name change). And I'm not really sure what to make out of the latter part of your comment @chrisarevir  😅

ben-burlingham:
@chrisarevir the "name" property was changed to "key" but the code hadn't caught up yet, it's just a fix for that.

PR #99, opened 2017-05-25
Turn off select all via prop

nickwmurphy:
Added a `isSelectAll` prop to allow the select all functionality to be turned on/off.

Prevented the dropdown from closing when clicking in a non-clickable area in `dropdownHead`.


nickwmurphy:
Whoops on all the commits. Got a little tangled with git.

PR #113, opened 2017-06-05
Tatari: check for change before sending a payload

nickwmurphy:
Before, the payload was sent no matter what. This branch stores the current filter set in the state, and compares that to the payload about to be sent. If it is the same, it is not sent.

Filter objects were converted to arrays to be sorted and then strings to be easily compared.

I am fairly certain this could be improved, but I hope this is close!


filters/tatari/src/Tatari.jsx
```diff
@@ -193,10 +206,25 @@ export default class Tatari extends React.Component {
...
+    const newSelections = Object.keys(payload.filters).reduce((acc, key) => {
+      acc.push(...payload.filters[key]);
+      return acc;
+    }, []);
+
+    newFilters.push(...newSelections);
+
+    const isFiltersSame = newFilters.reduce((acc, key) => savedFilters.includes(key), true);
+    const isFiltersSameLength = savedFilters.length === newFilters.length;
+
```
nickwmurphy: Should everything here and above be broken into its own method?

filters/tatari/src/Tatari.jsx
```diff
@@ -193,10 +206,25 @@ export default class Tatari extends React.Component {
...
+    const { savedFilters } = this.state;
+    const newFilters = Object.keys(payload.filters);
+    const newSelections = Object.keys(payload.filters).reduce((acc, key) => {
+      acc.push(...payload.filters[key]);
+      return acc;
+    }, []);
+
+    newFilters.push(...newSelections);
+
+    const isFiltersSame = newFilters.reduce((acc, key) => savedFilters.includes(key), true);
```
chrisarevir: It isn't carrying the acc forward in this case, so this is prone to not work actually :/

filters/tatari/src/TatariDropdownCheckboxes.jsx
```diff
@@ -43,7 +44,10 @@ const TatariDropdownCheckboxes = ({
 
   const singleSelection = options.reduce((acc, option) => {
     if (count === 1 && option.checked) {
-      acc.push(<div className={styles.singleSelection}>{option.value}</div>);
+      acc.push(<div
+        className={styles.singleSelection}
+        key={`option-${option.key}`}
```
chrisarevir: nice!

PR #114, opened 2017-06-05
Generic Styling

nickwmurphy:
@ben-burlingham this branch removes the specific styling from all components. 

Should we replace `shared/variables.scss` with more generic colors, etc., or completely isolate styling to each component? This does the latter, but the former might make enforcing generic styling easier. <-- Yes, this was introduced as well

What do you and @chrisarevir think?


ben-burlingham:
`variables.scss` is a hold over from Procore, so that's something to consider. 

My vote would be for no shared styles between open source components whatsoever; they're sort of "faceless" and gray.  Components with business logic can compose the styles they need where they're used.

Not sure if I'm answering the question - TLDR "isolate styles to each component" sounds closer to the direction I had in mind but open to discussion.

filters/rapu/src/Rapu.jsx
```diff
@@ -1,5 +1,5 @@
 import React, { PropTypes } from 'react';
-import cx from 'classnames';
+import cx from 'classnames'; // eslint-disable-line
```
ben-burlingham: ESLint disabling is generally a smelly smell smell

...if `cx` is unused feel free to remove.

nickwmurphy:
I agree with the desire to keep them "faceless" and gray. My main question is: should we have a shared variables file of "gray" colors, etc. that all the components use? That way, it might be easier to keep them consistent? I don't think this is a big deal either way.

filters/rapu/src/Rapu.jsx
nickwmurphy: The linter says `'classnames' should be listed in the project's dependencies, not devDependencies. (import/no-extraneous-dependencies)` on that line. `cx` is used in this file.

ben-burlingham:
Oh I see. Yeah sure that sounds good!

filters/rapu/src/Rapu.jsx
ben-burlingham: Oh in that case:

`yarn remove classnames` (maybe ` --save-dev`)
`yarn add classnames` should remove it from dev deps and add it to regular

filters/rapu/package.json
```diff
@@ -19,8 +19,9 @@
   },
   "homepage": "https://github.com/chrisarevir/component-library#readme",
   "dependencies": {
+    "prop-types": "^15.5.10",
     "qs": "^6.3.0",
-    "react": "^0.14.3"
+    "react": "^15.5.4"
```
ben-burlingham: Wow!  Hope this works... :crossed_fingers:

PR #123, opened 2017-06-09
Adding in bulk editor components

brianyan:
Added in bulkEditorDropdown and bulkEditorComponent to list-partial

bulkEditorDropdown will be the `Tipako` component while bulkEditorComponent will be the `bulk-editor` component. 

I'm not sure about if sticking bulkEditorComponent to the notifications div is the right way to do it, let me know if you have other approaches.


ben-burlingham:
Approve FWIW but @chrisarevir is the expert here...

chrisarevir:
Perhaps we want that to be a more generic name since it won't always be an export dropdown. Something like controlComponentOne, controlComponentTwo

layouts/list-partial/src/ListPartial.jsx
```diff
@@ -34,6 +39,7 @@ const ListPartial = ({
 
       <div className={styles.notifications}>
         { notifications && notifications }
+        { bulkEditorComponent && bulkEditorComponent }
```
chrisarevir: This can be moved into it's own area right below notifications, since they won't be showing up at the same time either way

brianyan: changes made!

PR #140, opened 2017-06-14
Added in the ability to clear inputs

chrisarevir:


pickers/maramataka/src/Maramataka.scss
```diff
@@ -61,13 +92,18 @@
   }
 }
 
+.input::-webkit-inner-spin-button,
+.input::-webkit-outer-spin-button {
+  appearance: none;
+  margin: 0;
+}
```
ben-burlingham: Recommend using a `pattern` attribute to match `/\d/`.  Will solve negative years also.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-pattern

PR #141, opened 2017-06-15
Ripanga null references

brianyan:
Added in safe checks so that we are not referencing a null value.


tables/ripanga/README.md
```diff
@@ -4,6 +4,10 @@ A table component with self-managed checkboxes, sticky head, sticky right panel,
 
 # Change Log
 
+##### 2.2.4
+
```
ben-burlingham: oooooooooooo

nickwmurphy: yikes though

chrisarevir: omg why

PR #142, opened 2017-06-15
Ripanga Sidebar Sticky Head Cell Bug Fix

nickwmurphy:
**New comments:**

This branch fixes the problem where resizing the page caused the sidebar header to lose its sticky positioning.

I have decided to not try to fix the problem for the main table header in this branch. I do not think I can make a graceful, timely fix. Instead, this branch fixes the same problem in the table's sidebar header (just the one cell).

This is very similar to what needs to happen in the main header, but minus some of the complexities. 

The `resize` listener does not seem to have any affect at the moment, so replacing it with a method that is called just once to set the initial top value of the header seems appropriate. 

Also, one of the reasons this was easy to fix in the sidebar header and not the main header is the sidebar is a "smart" component. Maybe the header should be converted to one too? It would benefit from having its own lifecycle methods.

__________________________________________________________________________________________________________

**Old comments:**

This is a hack at the moment. Looking for some advice on the best way to do this. 

Problem: on table resize, `headerInitialTop` currently gets set. This is a problem because that value should only be set once (on initial page load). That value is what determines if `moveHeader` or `restoreHeader` is called to keep the header "sticky."

By reseting this value every resize, the value is no longer "initial" as intended, and the sticky header becomes buggy.

I created a global `flag` and wrapped the part of `resize` that sets `headerInitialTop` in a conditional statement, so as to only call it once. This fixes the problem, but it is certainly a hack.

Instead, I'd like to find a spot in the initial set up of the page to set this value.


ben-burlingham:
> This is a problem because that value should only be set once (on initial page load)

Not exactly true - if the page resizes, the header reference point (the offset) should be recalculated.

`headerInitial` is misleading - `headerOffset` or something would be better probably.

nickwmurphy:
I think resize is only an issue with the width of the table. It appears to me that calculating the header's top value just once is a good solution. Certainly open to talk about that. 

I think certain resize events (ones that push the table up or down) would require a recalculation of the header's top, but recalculating on every resize event is definitely causing some issues.

For example, if the header is at the top of the page and a resize event happens, its top value will be zero if calculated at that moment. That means the `restoreHeader` event will fire instead of the `moveHeader` event. 

  `if (scrollTop > headerInitialTop) {
      moveHeader(this.header, scrollTop - headerInitialTop);
    } else {
      restoreHeader(this.header);
    }`

`restoreHeader` sends it back to its original spot, whereas `moveHeader` keeps it "sticky." Only calculating the header's top once prevents this issue.

ben-burlingham:
Cool, thanks for investigating! Sounds like you have a pretty good understanding of what's going on so let's go with that.  Really appreciate your digging in here!

tables/ripanga/src/Ripanga.jsx
```diff
@@ -169,14 +169,16 @@ export default class Ripanga extends React.Component {
...
-      initialWidth += 30;
+      initialWidth += 42;
     }
 
     const tableWidth = this.props.columnDefinitions
       .reduce((acc, def) => (def.hidden ? acc : acc + def.width), initialWidth);
 
     this.header.style.minWidth = `${tableWidth}px`;
 
+    console.log(tableWidth);
```
chrisarevir: pls remove

PR #149, opened 2017-06-16
Ripanga VERY sticky header

nickwmurphy:
The Ripanga headRow component is now a "class level" component so that it has access to its own lifecycle methods.

The Ripanga header is now sticky at all times!


tables/ripanga/src/Ripanga.jsx
```diff
@@ -309,23 +216,27 @@ export default class Ripanga extends React.Component {
...
-              onScrollTrack: this.onScrollTrack,
-              onSort,
-              showGroups,
-              showCheckboxes,
-              sortState,
-              styles
-            }) }
-          </div>
+          <RipangaHeadRow
+            className={styles.tableHead}
```
chrisarevir: You can actually move this down to the object you have spreaded below

nickwmurphy:
Cool! Yea I believe `debounce` is still being used in the same manner as before. It is just moved out of the top level Ripanaga component since it was only being used there for the header. 

As for `getBoundingClientRect()`, we are definitely decreasing number of times that is called by being more selective about when `headerInitialTop` is calculated!

PR #152, opened 2017-06-19
Removing div if not needed

chrisarevir:


pickers/tipako/src/Tipako.jsx
```diff
@@ -305,11 +305,14 @@ export default class Tipako extends React.Component {
...
-      {clearAll}
-    </div>);
+    const controls =
+      (selectAll || clearAll) ? (
+        <div className={this.styles.controls}>
+          {selectAll}
+          {spacer}
+          {clearAll}
+        </div>
+      ) : null;
```
ben-burlingham: I just remembered the `&&` trick the other day:

```(selectAll || clearAll) && (<Foo />)```

YMMV saves some typing

PR #156, opened 2017-06-20
Maramataka: Type resolution fix.

ben-burlingham:


pickers/maramataka/src/Maramataka.jsx
```diff
@@ -35,7 +35,7 @@ export default class Maramataka extends React.Component {
       errors: { day: false, month: false, year: false },
       expanded: false,
       selected: { day: null, month: null, year: null },
-      value: props.value
+      value: { day: props.value.day * 1, month: props.value.month * 1, year: props.value.year * 1 }
```
piercefrawley: woah! didn't know about this trick 👍

ben-burlingham: Yep! It's fast and effective. Only downside is it's a little bit obscure. 

I might start using `Number()` if I remember it, but looks like `*1` is pretty quick: http://phrogz.net/js/string_to_number.html

PR #160, opened 2017-06-21
Tipako Style Cleanup

nickwmurphy:
Changes: Tipako now opens on click/focus of the input.

Style updates to make it match Tatari.


pickers/tipako/README.md
```diff
@@ -4,6 +4,8 @@ A picker with support for group or individual selections.
 
 # Change Log
 
+##### 3.4.6
+- Style cleanup
```
ben-burlingham: Can you put 

> now opens on click/focus of the input

here so it's more descriptive to future us?

nickwmurphy: Yep!

nickwmurphy: Done

PR #164, opened 2017-06-23
Bulk Editor Empty State

nickwmurphy:
If there are no items being passed in, display an empty message and disable the update button.


layouts/bulk-editor/src/BulkEditor.jsx
```diff
@@ -73,6 +70,7 @@ BulkEditor.propTypes = {
 BulkEditor.defaultProps = {
   cancelText: 'Cancel',
   children: [],
+  emptyMessage: 'Empty',
```
ben-burlingham: nb Could be a touch more descriptive?

layouts/bulk-editor/src/BulkEditor.scss
```diff
@@ -18,6 +18,10 @@ $defaultSpacing: 5px;
   flex-direction: column;
 }
 
+.emptyMessage{
+  font-weight: bold;
+}
+
 .fields {
   display: flex;
   flex: 7; // This is to take up 70% of the width
```
ben-burlingham: nb Not part of your PR but this comment could be misleading if the other elements in the layout have `flex-basis`s that add up to more than 3

nickwmurphy: I do not know much about this. @chrisarevir?

PR #170, opened 2017-06-26
Tipako: Spinner fix.

ben-burlingham:


pickers/tipako/src/Tipako.scss
```diff
@@ -144,6 +144,11 @@ $h: 30px;
   background-position: center center;
   background-repeat: no-repeat;
   background-size: 16px 16px;
+
+  &:hover {
+    background-color: unset;
```
chrisarevir: Can't use this in IE11 :/ http://caniuse.com/#feat=css-unset-value

PR #171, opened 2017-06-26
Tiwae Bottom Slot, Ripanga Row Height

nickwmurphy:
Added a bottom slot to Tiwae for custom buttons, etc.

Added support in Ripanga for row height to be controlled.


tables/ripanga/src/RipangaBodyRow.jsx
```diff
@@ -49,8 +50,10 @@ const RipangaBodyRow = ({
     </div>);
   }
 
-  return (<div key={`row-${rowData[idKey]}`} className={styles.tableRow}>
-    {cells}
+  return (<div
+    key={`row-${rowData[idKey]}`}
+    className={cx(styles.tableRow, { [styles.minHeight]: isMinHeight })}
+  >{cells}
```
ben-burlingham: nb newline probably

tables/tiwae/sandbox/sandbox.jsx
```diff
@@ -46,17 +46,32 @@ const onChange = (newColumnDefinitions) => {
   console.warn(newColumnDefinitions); // eslint-disable-line
 };
 
+const onRowHeightChange = (evt) => {
+  evt.stopPropagation();
```
ben-burlingham: This seems awkward, will all implementers have to stopPropagation themselves?

Recommend pulling into component.

tables/tiwae/sandbox/sandbox.jsx
```diff
@@ -46,17 +46,32 @@ const onChange = (newColumnDefinitions) => {
...
+  className={sandboxStyles.rowHeightControl}
+  onClick={onRowHeightChange}
+>Change Row Height</div>);
+
 render(
   <ColumnOrganizer
     {...{
       columns,
       defaultColumns,
+      bottomSlot,
```
ben-burlingham: nb We use `slot` as a prefix in the other components, maybe we should stick with the same convention?

tables/tiwae/src/Tiwae.jsx
```diff
@@ -198,6 +202,10 @@ export default class Tiwae extends React.Component {
       <div className={styles.reset} onClick={this.onReset} >Reset to Default</div>
     </div>);
 
+    const slot = this.props.slot
+    ? <div className={styles.slot} onClick={this.onItemClick}>{this.props.slot}</div>
+    : null;
```
ben-burlingham: nb Can use `this.props.slot && <div ...`

PR #172, opened 2017-06-27
Tipako: Bottom Slot

nickwmurphy:
Added a bottom slot to Tipako for custom buttons, etc.


pickers/tipako/src/Tipako.jsx
```diff
@@ -345,6 +348,10 @@ export default class Tipako extends React.Component {
       ? <div className={this.styles.slot}>{titleSlot}</div>
       : null;
 
+    const buttonSlot = bottomSlot
+      ? <div className={this.styles.bottomSlot}>{bottomSlot}</div>
+      : null;
+
```
ben-burlingham: nb Can use `bottomSlot && <div className={this.styles.bottomSlot}>{bottomSlot}</div>` to save some characters and improve readability a bit.

Can then maybe use that statement to plug in directly instead of assigning to `buttonSlot`

pickers/tipako/src/Tipako.scss
```diff
@@ -172,7 +172,7 @@ $h: 30px;
     background: #fff;
     border: 1px solid #939393;
     height: auto;
-    max-height: 300px;
+    max-height: 340px;
```
ben-burlingham: nb Have you considered conditionally adding a class to extend the `max-height` if the bottom slot is populated? Not sure if that's preferred, just food for thought.

pickers/tipako/sandbox/sandbox.jsx
```diff
@@ -75,15 +75,18 @@ const valueField = 'value';
 const titleValue = 'Calypso';
 const titlePlaceholder = 'Make a selection...';
 const titleSlot = <div className={`fa fa-bus ${sandboxStyles.titleIcon}`} />;
+const bottomSlot = <div className={sandboxStyles.bottomSlot}>Bottom Slot</div>;
```
ben-burlingham: nb Ripa (tabs) uses the pattern `slotL` and `slotR`.  I'd like to have a consistent pattern across the particles, maybe we could work together to decide on a convention?

pickers/tipako/sandbox/sandbox.jsx
```diff
@@ -75,25 +75,25 @@ const valueField = 'value';
 const titleValue = 'Calypso';
 const titlePlaceholder = 'Make a selection...';
 const titleSlot = <div className={`fa fa-bus ${sandboxStyles.titleIcon}`} />;
```
ben-burlingham: Should we change this while we're in there?

nickwmurphy: Good catch. I'll get that before merging.

PR #175, opened 2017-06-30
Wharangi: removed ellipsis from navigation under 4 pages

amanda-bosson:


navigation/wharangi/src/Wharangi.jsx
```diff
@@ -102,29 +102,29 @@ export default class Wharangi extends React.Component {
...
-    const prev = (active > 2)
+    const prev = (active > 1)
       ? (<button
         onClick={() => this.handleSelect(active - 1)}
         className={cx('fa', 'fa-angle-left', this.styles.prev)}
       />)
       : null;
 
-    const prevEllipsis = (active > 2)
+    const prevEllipsis = (active > 2 && (totalPages !== 3))
```
ben-burlingham: Any reason we just don't say `active > 3` here and below?

amanda-bosson: If `totalPages` is greater than 3, we want the previous ellipsis to show up starting on page 3. This just prevents the ellipsis from showing up on page 3 if that is the last page.

PR #178, opened 2017-07-06
Tipako: null titleValue handling.

ben-burlingham:


pickers/tipako/src/Tipako.jsx
```diff
@@ -359,7 +359,7 @@ export default class Tipako extends React.Component {
           placeholder={titlePlaceholder}
           ref={(input) => { this.searchInput = input; }}
           type='text'
-          value={value}
+          value={value || ''}
```
chrisarevir: I think this can be better handled as a default prop, no?

chrisarevir: Nvm, undefined !== null, different cases :/

ben-burlingham: Good thinking but looks like default prop is already set to `''`.

The issue is if the prop is passed as `null` - then the default prop never gets triggered, but the input still has a value of `null`.

PR #179, opened 2017-07-06
Add Lerna and Filters

ksespinola:
### Changes
- Introduce `lerna` for managing npm packages
- Use neutrino to build projects and generate sandboxs
- Add `@procore/filters` port of wrench shared filters


chrisarevir:
Discussion point: What benefit does adding Lerna provide?

Introducing global dependencies and patterns, like lerna, are major decisions that should probably be discussed with implementers.

Background: We have tried it in the past hoping it would solve some problems, but it turned out to be a buggy experience when installing it on other peoples computers, having it publish correctly, and having weirdness with permission handling (publishing). Introducing it to manage a subsection of the repo for just the things you and your team team will manage would be the ideal way to use it atm, but again, I don't see the benefit just yet and would love to hear the reasons!

Other discussion point: Any chance we could compromise and version up Tatari? Or is the use case different/the API not ready for it?

ben-burlingham:
Agreed - Lerna would be a significant global change so we should probably get together and decide if that's a complexity that we're interested in layering on.  

---

Good to see some Wrench components coming in! :) Also to Christian's point, Tatari uses the Filterable gem pattern from Procore's Vapid endpoints per Derek Carter's recommendation.

remyyounes:
Closing for: https://github.com/procore/particles/pull/198

PR #180, opened 2017-07-07
Tipako Key Listeners

nickwmurphy:
To fix:

- [x] Needs to work when `searchable = false`
- [x] Needs to work when the component is used multiple times in a layout
- [x] Listeners need to be pulled into their own methods to reduce complexity (see Ben's comment below) -- per conversation with @ben-burlingham, we are going to follow up on this.


pickers/tipako/src/Tipako.jsx
```diff
@@ -258,29 +275,45 @@ export default class Tipako extends React.Component {
...
+          const focusedItem = currentIndex === counter;
+
+          const childItem = result.concat(<div
             className={cx(this.styles.item, this.styles.childItem,
-              { [this.styles.disabled]: vv.disabled })}
+              { [this.styles.disabled]: vv.disabled,
+                [this.styles.keyFocus]: focusedItem })}
             key={`child-${v[keyField]}-${vv[keyField]}`}
+            onClick={(e) => { this.onChildClick(e, vv); }}
+            onKeyDown={(e) => { if (focusedItem && e.keyCode === 13) this.onChildClick(e, vv); }}
```
ben-burlingham: nb eek, ambiguity - harder to maintain and extend. Recommend pulling these (and any other) listeners into their own methods

nickwmurphy: I agree 100%. I was having some issues when not doing them inline, but I am going to give it another shot.

PR #189, opened 2017-07-12
Modal Component

piercefrawley:
## Changes
- Adds modal component for initial release
- Accepts basic show value, header, body, and footer as props

screen shot 2017-07-12 at 9 37 15 am


layouts/matapihi/src/index.js
```diff
@@ -0,0 +1,40 @@
...
+        </button>
+      </div>
+    );
+  }
+}
+
+render(
+  React.createElement(Component),
+  document.getElementById('root')
+);
```
chrisarevir: Interesting, I don't think you need this to explicitly render the modal, but rather just export it?

chrisarevir: More like:

```javascript

import Modal from './Modal';

export default Modal;
```

chrisarevir: If this logic is needed for the modal to work , I would suggest moving it back to Modal.jsx

chrisarevir: Ah, this looks like sandbox copypasta'd over. A mistake perhaps?

Marswang92: Oh my bad that's for dev use. I'll fix rn. Thanks!

PR #202, opened 2017-07-17
Give option of headerless and footerless

Marswang92:
# Change Log
- Header and footer will not be rendered if passed in `disableHeader` and `disableFooter` props.
- Close button should be rendered differently when there's no header

# Before
screen shot 2017-07-17 at 12 00 39 pm

# After
screen shot 2017-07-17 at 12 00 54 pm


chrisarevir:
I think the default option should render a header, having to pass it in everywhere seems a little odd for the general use case of a modal popup

Marswang92:
@chrisarevir Gotcha that make sense! So what I'm gonna do is giving a prop `disableHeader`, does that sound good?

PR #203, opened 2017-07-17
Ripanga: Minor Style Fixes

nickwmurphy:


tables/ripanga/src/Ripanga.scss
```diff
@@ -56,7 +56,7 @@ $tableZ: 0;
   border-width: 0 0 1px 0;
   display: flex;
   flex-grow: 1;
-  padding: $padding;
+  padding: 0 8px;
```
chrisarevir: Why don't you use the $padding variable?

nickwmurphy: you are cruel

PR #208, opened 2017-07-24
Location Picker 2.0.0

nickwmurphy:
Using the same data management strategy as version 1.0.1, Tauwahi 2.0.0 will allow for a tiered data structure to be navigated, selected, and created. 

This experience is no longer optimized for injection into a modal. It has been optimized for injection into the dropdown region of a picker (like Tipako), but it can stand alone and is, for the most part, responsive. 

See how it works with Tipako here: http://tiered-picker.herokuapp.com/

See the Tipako-Tauwahi implementation here: https://github.com/nickwmurphy/tiered-picker/blob/master/src/TieredPicker.jsx

**Navigate:**
screen shot 2017-07-30 at 11 13 52 am

**Select:**
screen shot 2017-07-30 at 11 51 39 am

**Create:**
screen shot 2017-07-30 at 11 51 13 am


pickers/tauwahi/src/Tauwahi.jsx
```diff
@@ -36,150 +37,329 @@ export default class Tauwahi extends React.Component {
...
   }
 
-  onCreate = () => {
-    const names = this.state.stack.map(id => this.memoizedData[id].name);
-    this.props.onCreate(names);
+  onSelect = () => {
+    const { stack } = this.state;
+    const selection = this.memoizedData[stack[stack.length - 1]];
+
+    console.log(selection);
```
nickwmurphy: Fixed 👍

PR #210, opened 2017-07-27
Marmataka React 16 Compatibility

piercefrawley:
## Change Log

React 16 isn't playing nicely with [setState async callbacks](https://github.com/facebook/react/issues/10294) so we had to update them


chrisarevir:
Maramataka no longer exists, we can re-visit this when 16 becomes a real thing!

PR #211, opened 2017-07-28
Tipako: dropdownContent prop and titleValue Bug Fix

nickwmurphy:
Fixed a bug around the `titleValue` prop. The title value was always updating to the original prop that was passed after clearing and blur.

Added a prop called `dropdownContent` that allows a node to be passed in and displayed instead of the typical item list.


pickers/tipako/src/Tipako.jsx
```diff
@@ -169,10 +177,16 @@ export default class Tipako extends React.Component {
     this.updateValue(item);
   }
 
-  onCaretClick = () => {
+  onCaretClick = (evt) => {
```
chrisarevir: Note for later: rename to convey actual usage

pickers/tipako/src/Tipako.jsx
```diff
@@ -460,10 +467,15 @@ export default class Tipako extends React.Component {
           className={cx(this.styles.staticText, { [this.styles.noClear]: !updateOnSelect })}
           onClick={this.onCaretClick}
         >
-          {(value.length > 0 && value) || titleValue || titlePlaceholder}
+          {value || titlePlaceholder}
         </div>
       );
 
+    const itemsContainer =
+      !dropdownContent && (<div className={this.styles.itemsContainer}>
```
chrisarevir: Spacing!

PR #228, opened 2017-08-08
Ripanga: z-index updates, removing default styles.

ben-burlingham:


tables/ripanga/src/Ripanga.scss
```diff
@@ -166,26 +161,25 @@ $tableZ: 0;
...
+  // z-index: 0;
+  //
+  // &.focused {
+  //   position: fixed;
+  //   z-index: 1;
+  // }
+  //
+  // &:active {
+  //   outline: 0 !important;
+  // }
```
chrisarevir: Not needed?

PR #234, opened 2017-08-09
Maramataka to DateSelect Rename

nickwmurphy:


date-select/package.json
```diff
@@ -1,23 +1,23 @@
 {
-  "name": "maramataka",
+  "name": "date-select",
```
chrisarevir: `@particles/date-select`

PR #264, opened 2017-08-15
Table Alignment and Last Column Overlap Bug Fixes

nickwmurphy:
1. Fixed a bug where the header was not aligning with the body rows.

2. Fixed a bug where the sidebar container was hiding the last 26px of the last column.


table/src/HeadRow.jsx
```diff
@@ -87,14 +87,8 @@ export default class HeadRow extends React.Component {
...
-
-    if (showGroups) {
-      initialWidth += 30;
-    }
-
-    if (this.props.showCheckboxes) {
-      initialWidth += 42;
-    }
+    if (showGroups) initialWidth += 16;
+    if (this.props.showCheckboxes) initialWidth += 42;
```
nickwmurphy: We need to find a better way to do this (here and above) in my opinion. Second time it has caused an issue.

ben-burlingham: Agreed.

table/src/Table.scss
```diff
@@ -7,7 +7,7 @@ $tableZ: 0;
 .contentContainer {
   border: 1px solid #c3c2c2;
   border-width: 0 1px;
-  padding-right: 100px;
+  padding-right: 126px;
```
ben-burlingham: I'm not convinced this solves the issue.  It feels a lot like the view used for reference has 13 columns with 2px of border...or some other specific use case.  Can we be reasonably sure this will fix all use cases?

nickwmurphy: This is actually because the sidebar container is 128px in width. Then 2 was subtracted because of some borders. I could use some variables in the SCSS to make that more explicit.

PR #270, opened 2017-08-17
Select: Test data update.

ben-burlingham:
It wasn't clear that `titleValue` could actually be any value, not one in the `data` collection.


piercefrawley:
thanks Ben!

PR #282, opened 2017-08-22
Make inline select searchable for table

Marswang92:
# Change Log
- Add a `searchable` props for inline select.
- Use the search bar to filter dropdown list according to its value.

gaz6rkmy8b


table/sandbox/sandbox.jsx
```diff
@@ -33,6 +33,7 @@ const renderCell = (rowData, columnDef) => {
         { key: 11, value: 'Burmese Bamboo Shark' },
         { key: 12, value: 'Megamouth Shark' }
       ]}
+      searchable={true}
```
ben-burlingham: nb No need for `true`, it's implicit

table/src/InlineSelect.scss
```diff
@@ -97,6 +97,14 @@
   z-index: 1;
 }
 
+.searchInput {
+  width: calc(100% - 10px);
```
ben-burlingham: Avoid `calc` at all times, bad perf and almost always unnecessary

table/src/InlineSelect.scss
```diff
@@ -97,6 +97,14 @@
...
+  padding: 3px;
+  margin: 5px;
+  border-radius: 2px;
+  border: 1px solid $gray;
+}
```
ben-burlingham: Please alphabetize CSS props

table/src/InlineSelect.jsx
```diff
@@ -116,26 +109,55 @@ export default class InlineSelect extends React.Component {
...
-            {item.value}
-          </div>)
-        )
-      }
-    </div>
-  )
+  showDropdown = (e) => {
+    if (!e || e.target !== this.clear) {
+      this.setState({ expanded: true }, this.position);
+      document.addEventListener('click', this.hideDropdown);
```
ben-burlingham: This, and the listener below, are never removed.  This particle uses the `tabindex`/focus/blur pattern, not event listeners.  You'll have to use the existing `onMouseDown` pattern, which implicitly manages event listeners and performs well.  See `InlineDateSelect` for an example, or swing by my desk and I'll show you how.

Marswang92: It is removed when hideDropdown is triggered, so it will automatically clean itself.

PR #297, opened 2017-08-25
Grb/inline text class

yoyis3000:
Renamed the classes of InlineText to match the other Inline element's naming convention.


chrisarevir:
Inner/Outer container sound like they are the wrapping classes for the entire thing. Would it be possible to name them a little clearer?

yoyis3000:
@chrisarevir I can totally do that, the only reason I chose those class names because the Inline elements share those class names and because of that we only need one override.scss file.. to override them all. If I rename them I'd imagine you'd want me to rename them all? Which I don't mind doing, I just want to make sure that's the route we want to go.

table/README.md
```diff
@@ -4,6 +4,9 @@ A table component with self-managed checkboxes, sticky head, sticky right panel,
 
 # Change Log
 
+##### 4.4.0
```
ben-burlingham: nb `minor` semver is for new functionality; this is a breaking change so I'd probably recommend a `major` bump.

(`patch` is for nonbreaking changes that aren't new functionality)

yoyis3000: Ah, I wanted to make sure it was a major patch, and it was but I forgot to apply to the version numbers.

nickwmurphy:
This is going to clash with #300. Let's chat about this.

PR #300, opened 2017-08-25
Inline Editable Cell Cleanup

nickwmurphy:
All Inline Editable fields will now follow this convention:

inline-fields


table/README.md
```diff
@@ -4,6 +4,9 @@ A table component with self-managed checkboxes, sticky head, sticky right panel,
 
 # Change Log
 
+##### 4.3.1
```
ben-burlingham: These changes will break style overrides so IMO they should be `major`, not `patch`

ben-burlingham:
Can we extend these conventions to `date-select`, `select`, and `filters`?

table/README.md
nickwmurphy: I agree. I will bump to 5.0.0.

table/sandbox/sandbox.scss
```diff
@@ -1,5 +1,5 @@
 .sandboxContainer {
-  font: 11px sans-serif;
+  font: 11px Lato, sans-serif;
```
ben-burlingham: nb No Lato in sandbox

Issue #296, opened 2017-08-24
Unit Testing Framework


mcclayton:
Since the goal of these are to be heavily shared/reusable components, test coverage will be vital.
I'd absolutely love to help get tests in this repo and more than happy to take the lead on that.
Do we envision having a `yarn test` from within each particle component and co-located tests? Or a root level `yarn test` that runs tests for all components with tests in a root level folder?
Or some other option?

ben-burlingham:
Wonderful! Thank you.

There's no vision as of yet.  The test per particle sounds more in line with the architectural goals of this initiative, but both ideas are great.  We have three main concerns:

- Simplicity
- Speed
- Patterns that remain agnostic and support multiple frameworks

The `yarn test` plan is right on with all three of these - the implementing team can use whatever framework they wish, and just hook the tests to `yarn test`. Love it.  

I'd also like to make the test suite non-optional, but it doesn't need to be run on every `yarn dev`.  Perhaps we can work it into a `yarn publish` command, which prompts for a couple of things, runs the test suite, and preps a published stack on success.  I think @RyanFuentes  is working on a tiny solution for the publish command; we can chat in person if you like.

Thanks again! Really looking forward to this.  Also, would love to get some headless rendering benchmark solutions in place 🤔  but this is a stretch goal.

PR #302, opened 2017-08-25
[DMX-529] Adding Jest unit tests for search bar component.

mcclayton:
Ticket: https://procoretech.atlassian.net/browse/DMX-529
Adding some unit tests to get the ball rolling for:
https://github.com/procore/particles/issues/296


search-bar/src/SearchBar.jsx
```diff
@@ -58,7 +58,7 @@ export default class SearchBar extends React.Component {
           className={styles.input}
           type='text'
           placeholder={this.props.placeholder}
-          onSubmit={this.onSubmit}
+          onSubmit={this.handleSubmit}
```
mcclayton: Found this as a result of a test. Didn't appear to be an onSubmit class method.

ben-burlingham: 🏆  already producing results!

The other particles use an `onFoo` convention so we should probably change this to match.

search-bar/test/SearchBar.test.jsx
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
```
ben-burlingham: Discussion point - Is it best to make a separate directory for a single file?  Maybe having it in the root would be more visible, and encourage people to maintain it better?

ben-burlingham: Oooh - can we output the test results inside the sandbox, a la Jasmine?

search-bar/src/SearchBar.jsx
mcclayton: Ok great, I'll change this to `onSubmit` in order to follow the `onFoo` convention.

search-bar/test/SearchBar.test.jsx
mcclayton: Might be able to do something like that. Did you specifically want it to be the test results? Or do you think something like CodeCoverage report would be good?
I'll probably tackle it in a follow-up PR but it'd be super simple to include Istanbul for code coverage reporting.

mcclayton:
Running tests are as simple as changing directory into the SearchBar particle directory and running `yarn test`. Sample run/output included in screenshot:

screen shot 2017-08-28 at 9 17 09 am

search-bar/test/SearchBar.test.jsx
ben-burlingham: Oooh coverage.  The dream.  A converage report was outside the scope of my comment but since you brought it up, hell yeah.  Coverage + results in the sandbox would be amazing.

search-bar/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["latest", "react", "stage-0"]
+}
```
chrisarevir: Is stage-0 necessary? Also, since it didn't exist before, is there a benefit from doing it here as opposed to the webpack config?

search-bar/src/SearchBar.jsx
ben-burlingham: Thanks! Please be sure to run `npm version patch|minor|major` (is it breaking? major), update the `README.md`, and run `yarn prod` when everything's ready.

search-bar/test/SearchBar.test.jsx
ben-burlingham: ...as long as it's straightforward, maintainable, and an intern can understand it!

chrisarevir: I would caution that we shouldn't have it as directly a part of the sandbox as it takes time to run the tests. It's also part of the great experience that there is instant feedback without having to wait for an entire system to reload, so I would lean with just the `yarn test` for now.

search-bar/.babelrc
mcclayton: Yeah stage-0 is necessary because search-bar uses static class properties i.e. `static propTypes = {`
I put it in .babelrc because this is where Jest (babel-jest) looks.

PR #303, opened 2017-08-27
Tiered-Select Style Upgrades

nickwmurphy:
1. Switching to a "select" button (instead of select happening on blur) based on user feedback.

2. Added a horizontal transform animation to better orient the user to the tiered pattern.

3. Added a more built out dataset to the sandbox.


ben-burlingham:
nb Would it be worth having the iconography the same for the forward/back interaction in the GIF?  I.E. both arrows or both carets

nickwmurphy:
Good thought. I want to stick with this for now and revisit in the future.

PR #305, opened 2017-08-28
Webpack dev server fix on all particles.

ben-burlingham:
No need to publish; patches are only in sandbox.


tiered-select/README.md
```diff
@@ -4,18 +4,22 @@ A tiered select.
 
 # Change Log
 
+<<<<<<< HEAD
```
mcclayton: Merge tag still here.

bulk-editor/package.json
```diff
@@ -38,7 +38,7 @@
     "sass-loader": "^6.0.6",
     "style-loader": "^0.18.2",
     "webpack": "^3.4.1",
-    "webpack-dev-server": "^2.6.1",
```
mcclayton: Did you want to give this particle a `PATCH` version bump as well?

ben-burlingham: Right yeah I should have specified.  The bulk-editor had a separate issue that causes the sandbox to break so I was going to fix that in a separate branch, so there's not a broken release.  Is that cool?

mcclayton: 👍

PR #306, opened 2017-08-28
Yarn sandbox

ben-burlingham:
Experimental: This introduces a dependency on `yarn` which may not be practical.  I'm not a huge fan, but it does work neatly.  Interested in opinions.


tiered-select/README.md
```diff
@@ -4,18 +4,22 @@ A tiered select.
 
 # Change Log
 
+<<<<<<< HEAD
```
mcclayton: Another merge tag 😛

PR #307, opened 2017-08-28
Search bar code coverage reporting via Jest + Istanbul

mcclayton:
With Mocha, it's a bit of of finagling to get code coverage reporting through Istanbul - but with Jest, it's a breeze.

This enables users to run `yarn coverage` to generate and print a code coverage report to the console.
screen shot 2017-08-28 at 11 14 55 am

HTML version of the report is available from `/search-bar/coverage/lcov-report/index.html`
screen shot 2017-08-28 at 11 13 51 am

**Discussion Point**
We don't have to do this now, but we can even set up code coverage thresholds so that if a PR reduces code coverage too much - we fail the `yarn coverage` command. Something to mull over for future tooling.


README.md
```diff
@@ -28,7 +28,7 @@ Ship it.
 
 # Development
 
-To develop a component `FOO`, all you need is these commands:
+To develop a component `FOO`, all you need are these commands:
```
ben-burlingham: nb `is` are less letters...something to think about.

README.md
```diff
@@ -44,6 +44,10 @@ In the tool which is importing `FOO`, use `yarn link FOO` to take advantage of t
 
 Starts the appropriate test runner for `FOO`'s test suite and runs the corresponding test file(s) from within the `/test` directory.
 
+### `yarn coverage`
+
+Generates a code coverage report for `FOO` component.
```
ben-burlingham: Do we want to wrap this into the test command?  Is there a reason to separate the two

mcclayton: We can certainly do that, the only reason I put them in separate commands is for separation of concerns. Do we *always* want to generate a code coverage report for every test run?

ben-burlingham: Not sure of the long term vision, but my gut would be yes, because I think it would be abandoned (people would gradually stop using it) if it's hidden in another command.  I'd sort of expect `yarn test` to do all the things. 

This could be a factor in onboarding and adoption as well - "oh, all I have to do is yarn test" is a good selling point and one less thing to learn about.  Thoughts?

ben-burlingham: Also, lots of commands mean people have to learn them, and will gradually stop using them, so maintenance suffers and they may or may not work.  If there's one command, people will learn it and will maintain it.

mcclayton: All valid points - I'll join it so a single `yarn test` command will run the tests and print out the code coverage report. This will mean all future implementors of `yarn test` should ensure they both run the tests and generate coverage reports as well. 👍

PR #312, opened 2017-08-30
Enable customized stylesheets for Modal

Marswang92:
# Change Log
- Added `stylesheets` props for Modal.
- Trimmed code, added propTypes.
- Alphabetized class properties.


modal/sandbox/sandbox.jsx
```diff
@@ -13,22 +15,18 @@ class Component extends React.Component {
...
-          </Modal>
-        </div>
+        <Modal
+          closeOnOverlayClicked
+          show={this.state.show}
+          closeModal={() => this.setState({ show: false })}
+          size='large'
+          header={'Header'}
+          footer={(<div className={sandboxStyles.footer}>Footer</div>)}
+          onShow={show => this.setState({ show })}
```
ben-burlingham: For header, footer, and body slots we should follow the `slotFoo` naming convention.

modal/src/Modal.jsx
```diff
@@ -81,46 +80,68 @@ class Modal extends Component {
...
-              {children}
-            </div>
-            { !disableFooter && (
-              <footer className={styles.footer}>
-                {footer || 'footer'}
-              </footer>
-            )}
+          <div className={closeClassName}>
+            <i
+              className='fa fa-close'
```
ben-burlingham: Are these styles guaranteed to be available to all consumers of this particle?  (third party devs...)

Also, I'd recommend using the Unicode X symbol that we hopefully use everywhere else to delete/close things. Check search bar, filters, select.

modal/src/Modal.jsx
```diff
@@ -81,46 +80,68 @@ class Modal extends Component {
...
+              onClick={this.onCloseClicked}
+            />
           </div>
+          { !disableHeader &&
+            <header className={styles.header}>
+              {header || <h1>{label}</h1>}
```
ben-burlingham: Probably want to avoid `<h1>` here for semantic purposes and future accessibility concerns.  I recommend  a styled div.

modal/src/Modal.jsx
```diff
@@ -81,46 +80,68 @@ class Modal extends Component {
...
+Modal.defaultProps = {
+  stylesheets: [],
+  children: <div />,
+  footer: <div />,
+  header: '',
+  disableHeader: false,
+  disableFooter: false,
+  label: '',
+  size: 'medium',
+  onOverlayClicked: () => true,
```
ben-burlingham: What's the use case for exposing a click event anywhere on the modal?

And if this is not that, then we probably need a name that matches its functionality a bit better

modal/src/Modal.jsx
```diff
@@ -81,46 +80,68 @@ class Modal extends Component {
...
+  label: PropTypes.string,
+  size: PropTypes.string,
+  show: PropTypes.bool.isRequired,
   onOverlayClicked: PropTypes.func,
+  closeOnOverlayClicked: PropTypes.bool,
   beforeOpen: PropTypes.func,
-  afterOpen: PropTypes.func,
   beforeClose: PropTypes.func,
+  afterOpen: PropTypes.func,
   afterClose: PropTypes.func,
```
ben-burlingham: What are the use cases for exposing all of these before/after events?

modal/src/Modal.jsx
```diff
@@ -81,46 +80,68 @@ class Modal extends Component {
...
-  closeModal: PropTypes.func.isRequired,
-  closeOnOverlayClicked: PropTypes.bool,
-  title: PropTypes.string,
-  titleStyle: PropTypes.object,
-}
+  onCloseClicked: PropTypes.func,
```
ben-burlingham: `onOpen()` and `onClose()` should probably suffice for all of the above events?

modal/src/Modal.jsx
```diff
@@ -81,46 +80,68 @@ class Modal extends Component {
...
+  closeModal: PropTypes.func
+};
+
+Modal.defaultProps = {
+  stylesheets: [],
+  children: <div />,
+  footer: <div />,
+  header: '',
+  disableHeader: false,
+  disableFooter: false,
```
ben-burlingham: Disabling header and footer should probably be the responsibility of the parent component.  They'll send in a disabled markup rather than the component guessing about it.

ben-burlingham:
@mcclayton up to you and Mars, I'm not authoring anything here so can't weight in much except for convention :)  can you guys sync up?

modal/src/Modal.jsx
Marswang92: So there're 2 situation `onClose` will be triggered. Either user click close button or user click on overlay. However these props are optional just in case some users might need it.

modal/src/Modal.jsx
Marswang92: Previously some people think header and footer should be there by default. The way I think the best is rendering header or footer according to these 2 props are fed in or not. If they're null then not render those.

modal/src/Modal.jsx
Marswang92: Overlay is the transparent background on the entire page, which is same as `handleClickOut`.

modal/src/Modal.jsx
Marswang92: I read some source code of other open source modal components and most of them provide these apis. They're there cuz people may wanna use them as `componentDidMount` and `componentWillUnmount` for the modal.

modal/src/Modal.jsx
Marswang92: For sure I'll work on that

modal/sandbox/sandbox.jsx
Marswang92: For sure will change the way of naming.

modal/src/Modal.jsx
Marswang92: I can change the icon in to unicode symbol. The `closeClassName` can be overwritten through the `stylesheets` props.

Marswang92:
@mcclayton man the modals looks awesome! For sure we should merge some patterns i.e. animation. The only thing is we're trying to introduce as less 3rd party dependencies as possible so we might not wanna use `react-portal`. I'm down to discuss with you how we wanna do this!

modal/src/Modal.jsx
ben-burlingham: 1 less dep FTW

modal/src/Modal.jsx
ben-burlingham: Yep, we can fire `onClose()` in either case.  The `closeOnClick` prop (or whatever it was called) will say whether it does that or not

mcclayton:
@Marswang92 Sure thing, writing our own Portal component to wrap the Modal should be very simple if we want that dependency to be removed. Anywho, more just sharing this modal for visibility. No need to do anything with it yet if you don't want to.

Marswang92:
@mcclayton Fosho shall we setup a time tmr discussing what features we wanna see in the Modal component? Think @ben-burlingham is right it's time to figure out the features for a Modal 2.0.

modal/src/Modal.jsx
ben-burlingham: If the use case isn't present, let's add the complexity later.

mcclayton:
Yeah absolutely! I'm free anytime after 2:00pm tomorrow

modal/src/Modal.jsx
ben-burlingham: Ok, but why would someone want to know if the overlay is clicked? Wouldn't it just close is the prop is set, and `onClose()` fire?

PR #316, opened 2017-08-31
Setup Storybook

mboperator:
## Problems we're trying to solve
Developer documentation - it'd be great to get a live demo along with logging on functions injected as props.

## Justification for storybook
- It's fairly unintrusive (2 folders at the root, and some changes to our package.json)
- It's easy to build and deploy (literally run yarn build-storybook, and upload the output to your favorite static asset host)

## What this does
- [x] Adds Storybook
- [x] Prevents storybook webpack config from processing built components

particle-storybook

## Next Steps
- [ ] Add all components
- [ ] Experiment with turning every sandbox into a `story`


mcclayton:
_Sort of irrelevant so feel free to ignore_. 
Really just thinking out loud here:

Not sure if it'd be an abuse of Storybook, but do you think it'd make sense to have modules something like:
```
.add('tests', () => {
   <TestResultRenderer />
});
```

Where within the storybook we could easily see test results & code coverage for each component?

Just asking because if that makes sense I might like to implement that if this gets merged.

Marswang92:
Are we planning to replace `sandbox` with `react-storybook`?

mboperator:
@mcclayton that'd be fantastic. I think we could definitely script it. 

I can see a scenario where circle creates a json keyed by component name with the test results as a value. Then it's as easy as consuming that in stories/index.js

mboperator:
@Marswang92 it's on the table. We're still exploring solutions.

I'm working on a branch that imports the sandbox in to story book now.

Ben is working on a solution that uses Jekyll.

mboperator:
Closing as #319 is preferable.

PR #320, opened 2017-09-01
Use jQuery height method to properly align sidebar rows with table ro…

scottwalstead:
…ws when the user zooms in and out

Bugfix - The table rows and sidebar rows would be a few pixels out of alignment when they zoom in and out due to our use of offsetHeight. This branch attempts to use jQuery if it's available on the window so we aren't importing another version of jQuery as a dependency and offsetHeight is a fallback. The 1px border on the header row caused a minor bug where the height of the row wasn't accurate for about 1/3 of the zoom levels so I moved the row border to the cell levels which resolved the issue and resulted in proper alignment for all view levels.

This is my first PR, please let me know if I've missed any steps or need to do something differently
I wasn't sure how to run linting or tests if there are any
Also, does QE directly test here in the sandbox? If so, I'm not sure how they will test this code with window.$


table/src/HeadRow.jsx
```diff
@@ -25,6 +25,10 @@ const restoreHeader = (el) => {
   });
 };
 
+// jQuery height method is more accurate and offers better browser
+// compatibility than offsetHeight, which can sometimes be off by a few pixels
+const getElementHeight = (el) => typeof $ !== 'undefined' ? `${$(el).height()}px` : `${el.offsetHeight}px`;
```
chrisarevir: Having a soft dependency on a third party library not mentioned anywhere smells like an anti-pattern

.gitignore
```diff
@@ -3,3 +3,5 @@ lerna-debug.log
 node_modules
 npm-debug.log
 yarn-error.log
+.idea/
+package-lock.json
```
chrisarevir: is this from npm 5?

chrisarevir: Protip for `.idea` and their like: 

You can add those to your personal global `.gitignore` so you don't have to add them to other repos.

scottwalstead: Thanks for the tip. Will do. And I'm running npm 5.3. Not sure if package-lock is a new feature of npm 5 or not.

table/src/HeadRow.jsx
scottwalstead: So Marcus and I had a conversation and thought that using the jQuery on the window present in procore/procore might be an option to avoid bringing in another jQuery version.

Had a follow up discussion with Ben today and he suggested looking at the jQuery source for how they are doing it. The source code was complicated and I found another solution using getBoundingClientRect().height that worked pretty well. Ben pointed out that performance might be an issue with that so I did performance testing and documented my findings in the code. Basically it's only really slow if you read / write in the same line of the for loop. If you put the read / write in separate loops, for 1000 rows it takes 4ms instead of 4 seconds.

scottwalstead:
Doing IE browser compatibility testing then I will be updating this PR

scottwalstead:
Had a follow up discussion with Ben today and he suggested looking at the jQuery source for how they are doing it instead of a soft dependency on jQuery. The source code was complicated and I found another solution using getBoundingClientRect().height that worked pretty well. Ben pointed out that performance might be an issue with that so I did performance testing and documented my findings in the code. Basically it's only really slow if you read / write in the same line of the for loop. If you put the read / write in separate loops, for 1000 rows it takes 4ms instead of 4 seconds.

Code is ready for another pass

PR #326, opened 2017-09-06
Added prop to optionally hide Table Sidebar

dancingshell:
### Summary
Needed a table without a sidebar. The default behavior of leaving the renderSidebar* props null left a blank column where the sidebar was expected to be. This was due I believe to the padding-right attribute on the main table container statically set to sidebar width even though sidebar had no content.

### Fix
* Added prop ~~`hideSidebar`~~ `showSidebar` to `<Table/>` component that must be added to table component in order to see sidebar
* Removed padding-right from `.contentContainer`
* ~~Added button on sandbox to test showing and hiding sidebar~~

### Test
Run sandbox for table and see table with and without sidebar rendering
* `cd table && yarn && yarn sandbox`
* open localhost:8081


ben-burlingham:
Ooh just noticed I said "defaulting to true" - this doesn't have to be the case. Actually, should probably default to false, and be opt-in.  Thoughts?

dancingshell:
@ben-burlingham defaulting to false would be breaking change right? So if we don't mind the major version bump... either way is ok with me

ben-burlingham:
Yep major is fine.

table/src/Sidebar.jsx
```diff
@@ -142,6 +146,6 @@ export default class TableSidebar extends React.Component {
       ref={(el) => { this.header = el; }}
     >{headCell}</div>);
 
-    return <div className={styles.sidebarContainer}>{rows}</div>;
+    return <div className={ classnames(styles.sidebarContainer, { [ styles.hidden ]: hideSidebar })}>{rows}</div>;
```
ben-burlingham: Rather than rendering and then hiding, maybe a better strat would be to not render at all if `showSidebar=false`

dancingshell: @ben-burlingham I like you suggestion but when I tried to implement I get weird cell layout where the sidebar would have been. Maybe you could stop by and check it out? Or I could just leave as it was

table/src/Sidebar.jsx
```diff
@@ -32,15 +33,17 @@ export default class TableSidebar extends React.Component {
...
   };
 
   static defaultProps = {
     collapsedIds: {},
     renderSidebarBodyCell: null,
     renderSidebarHeadCell: null,
-    renderSidebarGroupCell: null
-  }
+    renderSidebarGroupCell: null,
+    showSidebar: false
```
ben-burlingham: Making this prop `.isRequired` will remove the need for a default prop (it's being passed in either way, so may as well take advantage of that)

table/src/Sidebar.jsx
```diff
@@ -142,6 +146,6 @@ export default class TableSidebar extends React.Component {
       ref={(el) => { this.header = el; }}
     >{headCell}</div>);
 
-    return <div className={styles.sidebarContainer}>{rows}</div>;
+    return <div className={ classnames(styles.sidebarContainer, { [ styles.hidden ]: !showSidebar })}>{rows}</div>;
```
ben-burlingham: Rather than rendering each cell and then hiding it, perhaps the entire `<Sidebar>` component could just not be rendered? See comment on `Table.jsx`.

table/src/Table.jsx
```diff
@@ -258,16 +261,17 @@ export default class extends React.Component {
     return (<div className={styles.contentContainer}>
       <Sidebar
```
ben-burlingham: Here, I'd suggest adding a conditional like: 

```const sidebar = showSidebar ? <Sidebar ... /> : null```

then later

```
<div className={styles.contentContainer}>
    {sidebar}
    <div className={styles.tableContainer} ...
```

That way, the sidebar code will never even run if it's not needed (and the `.hidden` CSS can be removed)

File: table/README.md [hunk 5]
```diff
@@ -4,6 +4,11 @@ A table component with self-managed checkboxes, sticky head, sticky right panel,
 
 # Change Log
 
+##### 6.0.0
+- Added `showSidebar` prop (bool: default `false`) to `<Table />`.
+- Changes the default behavior of `<Table />`
+ which now hides the sidebar by default and only renders it via `showSidebar`.
+
 ##### 5.1.5
```
ben-burlingham: Might want to pull master, I think we're up to 5.1.8 now

table/sandbox/sandbox.jsx
```diff
@@ -78,17 +78,19 @@ render(
...
       Uncheck All
     </button>
-
     <Table
       globalKey='table-sandbox'
       idKey='key'
       onMounted={onMounted}
       scope='sandbox'
       showCheckboxes
+      showSidebar={ false }
```
ben-burlingham: nb Might not need the `false`. Similar to `showCheckboxes`, if the prop is there it's assumed to be true, if it's not, it's `false`. 

Since the sandbox is to illustrate the maximum capabilities of the Particle, we'd want to show the sidebar, just `showSidebar` (so it shows) will act the same as `showCheckboxes`.

table/README.md
```diff
@@ -4,6 +4,11 @@ A table component with self-managed checkboxes, sticky head, sticky right panel,
 
 # Change Log
 
+##### 6.0.0
+- Added `showSidebar` prop (bool: default `false`) to `<Table />`.
+- Changes the default behavior of `<Table />`
```
ben-burlingham: nb Might not need this second bullet point?  First one is pretty descriptive

table/README.md
```diff
@@ -4,6 +4,13 @@ A table component with self-managed checkboxes, sticky head, sticky right panel,
 
 # Change Log
 
+<<<<<<< HEAD
```
chrisarevir: Merge Conflict left here

table/src/Table.jsx
```diff
@@ -39,7 +39,8 @@ export default class extends React.Component {
     showCheckboxes: PropTypes.bool,
     sortState: PropTypes.shape(),
     stylesheets: PropTypes.arrayOf(PropTypes.shape()),
-    tableData: PropTypes.arrayOf(PropTypes.shape()).isRequired
+    tableData: PropTypes.arrayOf(PropTypes.shape()).isRequired,
+    showSidebar: PropTypes.bool
```
chrisarevir: Alphabetize please!

table/README.md
ben-burlingham: Nice catch

PR #328, opened 2017-09-06
Setup unit testing for Table InlineText/BodyRow Component

Marswang92:
# Description
- This story will be to set up unit tests for the shared Table component, following the pattern and interface for running tests from within particles.
- Only invoke `composeStyles` when necessary in `InlineText`.

# Tasks
- [x] Install dependencies for testing.
- [x] Unit Testing for `InlineText`.
- [x] Unit Testing for `BodyRow`.

~~- [ ] Unit Testing for `InlineSelect`.~~
~~- [ ] Unit Testing for `InlineDateSelect`.~~
~~- [ ] Unit Testing for `GroupRow`.~~
~~- [ ] Unit Testing for `BodyRows`.~~
~~- [ ] Unit Testing for `Caret`.~~
~~- [ ] Unit Testing for `HeadCell`.~~
~~- [ ] Unit Testing for `HeadRow`.~~
~~- [ ] Unit Testing for `SideBar`.~~
~~- [ ] Unit Testing for `Table`.~~
~~- [ ] (?) Snapshot testing.~~


table/src/InlineText.jsx
```diff
@@ -43,6 +38,12 @@ export default class InlineText extends React.Component {
     }
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (this.props.value !== nextProps.value) {
+      this.setState({ value: nextProps.value });
+    }
+  }
```
ben-burlingham: Merge master should fix these

table/package.json
```diff
@@ -6,7 +6,13 @@
   "scripts": {
     "dev": "NODE_ENV='development' webpack --progress --colors --watch",
     "prod": "NODE_ENV='production' webpack --progress --colors -p",
-    "sandbox": "NODE_ENV='development' webpack-dev-server --progress --colors --hot --inline ./sandbox/sandbox.jsx --config webpack.dev.config.js  --content-base ./../sandbox"
+    "sandbox": "NODE_ENV='development' webpack-dev-server --progress --colors --hot --inline ./sandbox/sandbox.jsx --config webpack.dev.config.js  --content-base ./../sandbox",
+    "test": "jest --runInBand --verbose"
```
mcclayton: Any reason we want to disable running tests in parallel via `--runInBand`?

table/test/InlineSelect.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+  );
+};
+
+let inlineTextWrapper;
+
+beforeEach(() => {
+  inlineTextWrapper = renderInlineEdit();
+});
+
+describe('', () => {
```
mcclayton: Typically the outermost `describe` would be the name of the component being tested. i.e. `describe('InlineSelect', ...`

table/test/InlineSelect.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+  describe('Constructor', () => {
+    it('initializes `value` state with `value.` prop', () => {
+      expect(inlineTextWrapper.state('value')).toEqual(inlineTextProps.value);
+    });
+  });
+  describe('Input', () => {
+    it('should show after clicking on the container.', () => {
```
mcclayton: The `it` statement here doesn't really describe well what the test is testing.
Perhaps this should be 
```
describe('Input Container', () => {
    describe('onClick', () => {
        it('should set `editing` state value to `true`.', () => {
```
Perhaps there is a better way to word the describes - but hopefully you get the idea.

table/test/InlineSelect.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+      inlineTextWrapper.find('div').at(0).simulate('click');
```
mcclayton: This is a pretty brittle selector. Can you use http://airbnb.io/enzyme/docs/api/ReactWrapper/ref.html to select this div via it's `ref` ?

table/test/InlineSelect.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+      expect(inlineTextWrapper.state('editing')).toEqual(true);
+    });
+    it('should show `textEmpty` props if no `value` provided.', () => {
+      const wrapperWithoutValue = renderInlineEdit({ value: '' });
+      expect(wrapperWithoutValue.find('div').at(2).text()).toEqual(inlineTextProps.textEmpty);
```
mcclayton: Again, is there a better selector you can use here? Perhaps by a className, ID, ref?
I try to avoid selectors that rely on ordering at all costs.

table/test/InlineSelect.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+    });
+    it('should update `value` state correctly on change.', () => {
+      inlineTextWrapper.find('div').at(0).simulate('click');
```
mcclayton: Better selector?

mcclayton: What does the click do? Is this to set some sort of state variable before testing the changeEvent?
It's best not to have other functionality in a unit test that is unrelated to what it is you are purely trying to unit test. If this is just to get the input to show up - it'd be better to manually set the state of the component before the test rather than triggering a click event to do this.

table/test/InlineSelect.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+      const inlineTextInput = inlineTextWrapper.find('input');
+      const expectedChangeValue = 'This is a changed search string';
+      const mockChangeEvent = {
+        target: { value: expectedChangeValue }
+      };
+      inlineTextInput.simulate('change', mockChangeEvent);
+      expect(inlineTextWrapper.state('value')).toEqual(expectedChangeValue);
+    });
+    it('should clear `value` state on click clear button.', () => {
+      inlineTextWrapper.find('button').simulate('click');
```
mcclayton: Button is "ok" here, but this could still be better. The moment another button is added to this component, this selector could potentially fail the test when it shouldn't.

table/test/InlineSelect.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+      expect(inlineTextWrapper.state('value')).toEqual('');
+    });
+    it('should call appropriate onSubmit handler on Enter key up with `value` state if `value` has been updated.', () => {
```
mcclayton: This test is doing a lot here - It would be better if you were able to manually set the `value` in the state via: http://airbnb.io/enzyme/docs/api/ReactWrapper/setState.html instead of relying on the change event to do this. This keeps your test purely a unit test by only testing a single unit of code per test.

Also, I think the `describes` and `it` statement can then be cleaned up to something like:
```
describe('InlineSelect', () => {
    describe('Input', () => {
        describe('On Enter Key Up', () => {
             it('should call appropriate onSubmit handler with appropriate `value`.', () => {
```

mcclayton:
**Thank you so much** for taking the time to write unit tests. This is seriously incredibly needed!
You rock! 🚀  💯

table/package.json
Marswang92: Oh previously I was trying to solve the issue of console log not being emitted in the terminal. Will remove it.

table/test/InlineSelect.test.jsx
Marswang92: OK gotcha, will do a `setState` instead of trying to mock a click!

table/test/InlineSelect.test.jsx
Marswang92: Yeah actully wanna discuss with you about some issue of using class selector with css-module

table/test/data/groupedData.js
```diff
@@ -0,0 +1,51 @@
+module.exports = [
```
mcclayton: Typically a test directory like this is called `/fixtures`.

table/test/BodyRow.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+    });
+    it('should be checked if prop `isChecked` is set to true.', () => {
+      expect(bodyRowWrapper.find('input').prop('checked'))
+        .toEqual(props.isChecked);
+    });
+    describe('On Change', () => {
+      it('should call appropriate onCheck handler with row key.', () => {
+        const onCheckSpy = sinon.spy();
+        const wrapperWithSpy = renderBodyRow({ onCheck: onCheckSpy });
+        const mockChangeEvent = {
```
mcclayton: You don't need the `mockChangeEvent` here do you?
`onChange` does not doing anything with the actual event so you can simply call `wrapperWithSpy.find('input').simulate('change')`.

table/test/BodyRow.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+  describe('CheckBox', () => {
+    it('should be rendered if prop `showCheckboxes` is set to true.', () => {
+      expect(bodyRowWrapper.find(`.${styles.controlCheckbox}`).exists()).toEqual(true);
+    });
+    it('should be checked if prop `isChecked` is set to true.', () => {
+      expect(bodyRowWrapper.find('input').prop('checked'))
+        .toEqual(props.isChecked);
+    });
+    describe('On Change', () => {
+      it('should call appropriate onCheck handler with row key.', () => {
```
mcclayton: Did you mean appropriate `row key value` maybe?

mcclayton:
To keep things smaller, I'd perhaps isolate this PR to just the `InlineText` and `BodyRow` components. You can add others in individual PRs :)

table/src/InlineText.jsx
```diff
@@ -10,36 +10,41 @@ export default class InlineText extends React.Component {
...
+      value: props.value
+    };
   }
 
-  componentWillReceiveProps(nextProps) {
-    if(this.props.value !== nextProps.value) {
-      this.setState({ value: nextProps.value })
+  componentDidMount() {
+    // Only do this in browser.
+    if (this.container.parentNode.offsetWidth < 75 && !jest) {
```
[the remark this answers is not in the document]
Marswang92: Doing this cuz the warning is bit annoying in testing, thought it should only be triggered while it's being rendered in browser. (btw not a big fan of the parent Node Idea, cuz in different use case the `InlineText`'s parentNode might vary).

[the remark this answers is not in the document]
Marswang92: Test env is a better idea or I can just mock the container, will think about it!

table/test/data/groupedData.js
Marswang92: gotcha thanks!

table/test/BodyRow.test.jsx
Marswang92: Yes haha removing it!

table/test/BodyRow.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+      expect(bodyRowWrapper.find(`.${styles.tableCell}`).length).toEqual(2);
+    });
+  });
+  describe('CheckBox', () => {
+    it('should be rendered if prop `showCheckboxes` is set to true.', () => {
+      expect(bodyRowWrapper.find(`.${styles.controlCheckbox}`).exists()).toEqual(true);
+    });
+    it('should be checked if prop `isChecked` is set to true.', () => {
+      expect(bodyRowWrapper.find('input').prop('checked'))
+        .toEqual(props.isChecked);
```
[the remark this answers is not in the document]
Marswang92: Yeah i think in this case the better way to do is instead using `bodyRowWrapper`, just make a new wrapper explicitely set the prop like `const checkedWrapper = renderBodyRowWrapper({ isChecked: true })`?

table/test/BodyRow.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+let bodyRowWrapper;
+
+beforeEach(() => {
+  bodyRowWrapper = renderBodyRow();
+});
+
+describe('BodyRow', () => {
+  describe('Cells', () => {
+    it('should hide if hidden in definition is set to true.', () => {
+      expect(bodyRowWrapper.find(`.${styles.tableCell}`).length).toEqual(2);
```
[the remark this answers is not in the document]
Marswang92: So if the `def` set to be hidden the cell will be null. So here is an `existence` check, though it should be more specific e.g. instead of checking the length of nodes, check the existence of a specific node with def hidden.

table/src/InlineText.jsx
```diff
@@ -10,36 +10,41 @@ export default class InlineText extends React.Component {
...
+    textEmpty: PropTypes.string,
     value: PropTypes.string
   };
 
   static defaultProps = {
     canClear: true,
     onClear: () => {},
-    stylesheets: [],
-    value: 'Input...'
+    stylesheets: undefined,
```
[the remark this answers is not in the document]
Marswang92: Doing this here is just to shut eslint off and to avoid calling `composeStyles` when `stylesheets` is not provided. And yes actually set it default to be an array and check isEmpty in the constructor is a better idea. Thanks!

Marswang92:
@mcclayton For sure, will update this pr accordingly! Thanks a lot for all the super helpful advice!

ben-burlingham:
Is there a reason the fixtures reproduce the test data? Can they just pull from the sandbox test data?

Marswang92:
@ben-burlingham for sure we can. Doing this cuz lot of times we change stuff in the sandbox data for developing new features. The fixture folder instead should keep the same?

ben-burlingham:
I'd say exactly the opposite - the data used in the sandbox _is_ test data, and is there intentionally to test empty cases, overflow cases, etc.

Also, we don't want to maintain two stubs ;)

table/test/BodyRow.test.jsx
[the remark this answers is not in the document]
mcclayton: @Marswang92 Yes I like that option much better 👍

mcclayton:
Hmmm, I think @Marswang92 has a point that I agree with:
My only hesitation to only trying to use a single fixture for both is that while fixtures used for the sandbox could make good tests in the unit tests' suite, the reverse is not necessarily true. I'd hate to preclude ourselves from valuable test case fixtures simply because they don't make sense in the context of the sandbox. Anywho, just my two cents (I'm fine with whatever)
¯\_(ツ)_/¯

table/test/BodyRow.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+beforeEach(() => {
+  bodyRowWrapper = renderBodyRow();
+});
+
+describe('BodyRow', () => {
+  describe('Cells', () => {
+    it('should not be rendered if hidden in definition is set to true.', () => {
+      expect(bodyRowWrapper.find(`.${styles.tableCell}`).length)
+        .toEqual(columnDefinitions.filter(o => o.hidden !== true).length);
```
mcclayton: Nice, this is much better 👍

mcclayton: Also, back to the fixtures, sometimes it's better to inline them, or split specific fixtures out into their respective cases such as:
```
/fixtures/hiddenColumns.json
```
That way, you don't have one large shared fixture that is prone to breaking multiple tests when you modify it. Better to tailor your data to individual tests when possible vs. attempting to tailor one large fixture to the entire test suite.
I think this is fine for now, but just for future reference.

ben-burlingham:
@mcclayton you're the lead on the testing initiative, let's do what you think is best :)

mcclayton:
We're all "leads" ❤️

table/test/BodyRow.test.jsx
Marswang92: Got it this is very helpful. I'm on the way!

PR #329, opened 2017-09-06
Table Column End-User Resizing

nickwmurphy:
## Description
This branch adds the ability for an end-user to **resize** the widths of each of the table's columns via the interaction shown below.

Column widths are stored in session storage, and it is easy to implement reseting all columns to default widths.

Refactored to ensure highest level props are not mutated!

table-column-resizing

## Implementation
Three new keys are now available for each object in `columnDefinitions`.

1. `resizable`: bool that determines whether that column can be resized by an end-user
2. `maxWidth`: number that is the `max-width` (in px) to which that column can be resized
3. `minWidth`: number that is the `min-width` (in px) to which that column can be resized

**Note**: the `width` key is unaffected. It is used as the initial width and the default width for columns with `resizable: true`.

**Note**: In the case that `minWidth` and `maxWidth` are not set by a dev, the fallbacks are 50 and 300, respectively.

## Known Issue(s)

1. In IE11, `verticalLine` being set to `height: 100vh` makes `tableBody` become scrollable, resulting in misalignment with `Sidebar`. I need to find a different way to set the height of `verticalLine`.

2. Ideally, the cursor does not change from `col-resize` during the entire resize interaction, but that is proving to be tricky.


table/src/Table.scss
```diff
@@ -13,6 +13,7 @@ $sidebarZ: 3;
 
   * {
     box-sizing: border-box;
+    user-select: none;
```
ben-burlingham: ❤️

table/src/Table.scss
```diff
@@ -68,14 +68,58 @@ $sidebarZ: 3;
...
-}
 
-.headCellLabel {
-  padding: 0 $padding;
+  &.pressed,
+  &.pressed * {
+    cursor: col-resize;
+
+    ~ .headCell,
+    ~ .headCell * {
```
ben-burlingham: nb Usual warnings apply about `*` universal selector

nickwmurphy:
@ben-burlingham, prod build is `89.7 kB` with these changes.

table/src/HeadCell.jsx
```diff
@@ -3,54 +3,150 @@ import PropTypes from 'prop-types';
...
+    const { columnWidth } = this.props;
+
+    this.setState({ pressed: true, startWidth: columnWidth, startX: evt.clientX });
+  }
+
+  onMouseMove = (evt) => {
+    const { pressed, startX } = this.state;
+
+    if (pressed) {
+      this.setState({ displacement: evt.clientX - startX }, this.moveVerticalLine());
```
dannyphillips: NB: maybe instead of `displacement` cal it `delta`?

dannyphillips: Or mention `x_displacement`

table/src/Table.jsx
```diff
@@ -65,13 +65,15 @@ export default class extends React.Component {
 
     const checkedIds = JSON.parse(sessionStorage.getItem(`${props.scope}/CHECKED`)) || {};
     const collapsedIds = JSON.parse(sessionStorage.getItem(`${props.scope}/COLLAPSED`)) || {};
+    const columnWidths = JSON.parse(sessionStorage.getItem(`${props.scope}/COLUMN_WIDTHS`)) || this.getColumnWidths();
```
dannyphillips: Given we already use sessionStorage, this is a good re-use of that pattern. We may wanna consider changing this up down the road to not be dependent on sessionStorage.

nickwmurphy: I totally agree. I think the eventual goal is to store to the db (like configurable columns are) so that user does not have to constantly set their preferred column widths.

table/src/HeadCell.jsx
nickwmurphy: I like `delta`. Maybe `deltaX` would be most descriptive?

PR #331, opened 2017-09-06
update this.style on this.props.stylesheets updating

evandapolarbear:
Im bad at spelling branch names.

## Change Log
- update this.styles in select with compose styles and this.props.styles on props updating


brock-noah:
`yarn prod` and commit the built files afaik

PR #334, opened 2017-09-07
Simplifying InlineSelect & InlineDateSelect Positioning

nickwmurphy:
## Description
This branch removes unnecessary complexity from the positioning methods of `InlineSelect` and `InlineDateSelect`. This does not fix our overflow problems, but it does give us some better options going forward. I am going to follow up with a Github Issue to outline the remaining challenges.


table/README.md
```diff
@@ -4,6 +4,9 @@ A table component with self-managed checkboxes, sticky head, sticky right panel,
 
 # Change Log
 
+##### 6.1.2
+- `InlineSelect` and `InlineDateSelect` overflow big fix
```
ben-burlingham: nb Rather than "big fix" can we say what was fixed

nickwmurphy: "Bug" whoops. But yes needs to be more descriptive.

PR #335, opened 2017-09-08
Show Sidebar Padding Patch

nickwmurphy:
Adding padding-right to `.contentContainer` when `showSidebar` is true so that final column is not covered.


table/src/Table.jsx
```diff
@@ -290,20 +291,20 @@ export default class extends React.Component {
...
-              collapsedIds,
-              idKey,
-              renderSidebarBodyCell,
-              renderSidebarHeadCell,
-              renderSidebarGroupCell,
-              showGroups,
-              tableData,
-              styles
-            }
+          ...{
```
chrisarevir: S P A C I N G

PR #336, opened 2017-09-08
sticky-footer initial commit

evandapolarbear:
### Sticky Footer

This component is a sticky footer that will respect the presence of a page footer. It will remain stuck to the bottom of the window until it encounters the page footer. When it encounters the page footer, it will move up to reveal it.

It "knows" it has encountered the page footer because the height of the page footer can be passed in as a prop.

#### props:
`bottom`: the height of the page footer, also the maximum 'bottom' value of `StickyFooter`,
`content`: the children/content of `StickyFooter`,
`stylesheets`: same as always

#### Size:
28.8kb


sticky-footer/README.md
```diff
@@ -0,0 +1,6 @@
+# Sticky footer
+
+A parent component for fixing children to bottom of the screen.
+
+##### 1.0.0
+- Initial Commit
```
nickwmurphy: `nb`: Using `###` in README markdown for major versions.

nickwmurphy:
Needs a `sandbox` and `package.json`.

sticky-footer/src/Footer.jsx
```diff
@@ -0,0 +1,86 @@
...
+    const { pushHeight, bottom, left, right } = this.props
+
+    this.state = {
+      windowHeight: window.innerHeight,
+      bodyHeight: Math.max(document.body.scrollHeight,
+                           document.documentElement.offsetHeight,
+                           document.documentElement.scrollHeight),
+      windowWidth: window.innerWidth,
+      pushHeight,
+      bottom,
```
mboperator: How do you feel about wrapping `bottom`, `left`, `right` in an `offsets` prop object? I think it'd be a little bit more descriptive

sticky-footer/src/Footer.jsx
```diff
@@ -0,0 +1,86 @@
...
+      left,
+      right,
```
mboperator: Copying props to state is an antipattern _unless_ they're used as initial values and are mutated later.

If those initial props aren't mutated by this component, they should remain as props. 

The reasoning is that this leads to confusing bugs (ie: the props are changing but my component isn't!) and it saves us from writing logic to copy the changed props into state in `componentWillReceiveProps`.

sticky-footer/src/Footer.jsx
```diff
@@ -0,0 +1,86 @@
...
+      bodyHeight: Math.max(document.body.scrollHeight,
+                           document.documentElement.offsetHeight,
+                           document.documentElement.scrollHeight)
+    }), 250);
+
+    document.addEventListener('scroll', this.handleScroll.bind(this), false);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('scroll', this.handleScroll.bind(this), false);
```
mboperator: This line is not properly removing the event listener, because `this.handleScroll.bind` returns a reference to a new function every time.

The solution here is to set `this.handleScroll = this.handleScroll.bind(this)` in the `constructor`. Then you just use `this.handleScroll` in add and remove event listener.

sticky-footer/src/Footer.jsx
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  handleScroll() {
+    let amtToPush = this.calcPush();
+
+    if( this.state.bottom !== amtToPush ) {
+      amtToPush = amtToPush > 0 ? amtToPush : 0;
```
mboperator: You can use `Math.max(amtToPush, 0)` here

sticky-footer/src/Footer.jsx
```diff
@@ -0,0 +1,86 @@
...
+      windowWidth: window.innerWidth,
+      pushHeight,
+      bottom,
+      left,
+      right,
+    }
+  }
+
+  componentDidMount() {
+    setTimeout(this.setState({
```
mboperator: This isn't waiting. This is being called immediately. If you want setTimeout to wait before setting state you want to wrap it in an anonymous function, or use bind:
```js
setTimeout(() => this.setState({ ... stuff }), 250)
setTimeout(this.setState.bind(this, { ... stuff }), 250)
```

ben-burlingham:
Marcus's comments are all great input and correct.  Looks like the component was re-done though, with those in mind. Thanks guys!

nickwmurphy:
All already addressed 👍

nickwmurphy:
Just completed all of those :+1:

PR #337, opened 2017-09-08
SearchBar: Analytics

ben-burlingham:
A possible analytics pattern for particles, testable and extensible.  If we like it, I'll upgrade other particles ASAP.

Pattern for particle `FOO`:
- `onAnalytics` optional prop, callback
- `scope` optional prop, handles multiple instances

When called, `onAnalytics` is passed `{ event, particle, scope }`
- `event` is the event (enumerated, see below)
- `particle` is the name of the particle

`FOO_ANALYTICS` enum
- Is `export`ed 
- Contains events such as `{ SUBMIT: 'submit', CLEAR: 'clear', ... }`
- Contains particle name: `{ ... NAME: 'Foo' }`

Here's the sandbox in action:

screen shot 2017-09-07 at 6 59 41 pm


search-bar/sandbox/sandbox.jsx
```diff
@@ -3,17 +3,31 @@ import React from 'react';
...
 
-import SearchBar from '../src';
+import SearchBar, { SEARCH_BAR_ANALYTICS } from '../src';
 
 const onSubmit = (value) => {
   console.warn('Submitted:', value); // eslint-disable-line
 };
 
+const onAnalytics = (obj) => {
+    console.warn('Analytics event: ', obj);
```
mcclayton: `nb`: Super minor, but some of your indentation in this PR seems to be at 4 spaces instead of 2.

search-bar/test/SearchBar.test.jsx
```diff
@@ -94,4 +94,35 @@ describe('SearchBar', () => {
       expect(searchBarWrapper.state('value')).toEqual('');
     });
   });
+
+  describe('Analytics', () => {
```
mcclayton: Tessssssts 🔥 🚀 ❤️

search-bar/src/SearchBar.jsx
```diff
@@ -35,6 +42,14 @@ export default class SearchBar extends React.Component {
     this.setState({ value });
   }
 
+  onAnalytics = (event) => {
+      this.props.onAnalytics({
```
mcclayton: *Discussion:* Do we want to standardize the shape of an analytics event somewhere? Or is it desirable for analytics events to be freeform. 
What's the end goal use of these analytics? If it's for a dashboard - having a consistent data model may help.

mcclayton: If we want to ensure consistency of shape/functionality, we could probably create a HOC that could wrap our components with the added benefit that we can clean up and consolidate analytics logic for separation of concerns. Just some thoughts.

ben-burlingham: Love it. Will explore and possibly do in follow up.

PR #338, opened 2017-09-08
slight tweak for group sidebar zindexes

RyanFuentes:
Fixes this issues below by giving the headSidebarCell a higher z-index than the groupSidebarCell

image


table/src/Table.scss
```diff
@@ -207,17 +207,18 @@ $sidebarZ: 3;
...
   background: #e7e7e7;
   border-top-width: 1px;
-  z-index: 1;
+  z-index: 30;
 }
 
 .groupSidebarCell {
   composes: sidebarCell;
   background: #f2f2f2;
+  z-index: 20;
```
ben-burlingham: nb Is there a reason for the 10x multiplier?  Seems like an accomodation for poor z-index management.

Could just use `0, 1, 2`?

PR #340, opened 2017-09-08
Create first install script

mboperator:
## What this changes
Current workflow:
```
yarn
cd <my-component>
yarn
```
With this change:
```
yarn
```

## Why
To improve developer experience in combination with #319

With the introduction of #319 it's inconvenient to go into **every** component folder and `yarn` before being able to run `yarn storybook` at the root.

## Screenshot
screenshot from 2017-09-08 15-29-21

## Reservations
But this is just `lerna bootstrap`! Yes. This is a small change to improve the experience before we have a larger conversation about `lerna`.
https://github.com/lerna/lerna

## Heads up
**This script requires at least Node version 8.0 (current stable build is 8.4.0)**


init.js
```diff
@@ -0,0 +1,36 @@
+const fs = require('fs')
+const util = require('util')
+const exec = util.promisify(require('child_process').exec)
```
mcclayton: Not sure if this is a big deal, but worth noting that using `promisify` will require Node 8. 
Is this something we actually want when it'd be pretty easy to make a `promisify` util ourselves.

Regardless, if we go with this, we should document our version requirements in the global readme or introduce an `.nvmrc` file or both.
cc @mboperator @ben-burlingham

ryanyogan: I wish just going to point that out, we can do a node lock however it may make more sense to not force into node 8+ 

I'll leave that to the smarter people in this PR to figure out.

mboperator: @mcclayton It is easy to make a `promisify` util ourselves, but I figured if it's included in Node we might as well use it.

I noted that Node version 8 is required in the **How do I get started?** portion of the README as step 0. Switching to Node 8 does cause issues with `node-sass` (as per usual). I think the safest way forward is probably to write a shim, so I'll do that in another branch.

mboperator: @mcclayton https://github.com/procore/particles/pull/359

ben-burlingham: Or, we could remove the `promisify` requirement

ben-burlingham:
Rather than requiring our users (who may not be engineers) to all debug and upgrade their node installations, is there a more backwards-compatible solution we could use?

Either, write vanilla promises, or use an existing NPM package, or...lots of options here that don't block people.

PR #343, opened 2017-09-09
InlineText: Cannot submit an empty string if canClear is false.

yoyis3000:
Checked if `value` is empty and `canClear` is false before submitting through `onBlur` and `onKeyDown`.


table/src/InlineText.jsx
```diff
@@ -58,16 +58,16 @@ export default class InlineText extends React.Component {
...
   }
 
   onSubmit = () => {
     const { previousValue, value } = this.state;
-    this.setState({ value, editing: false });
-    if (value !== previousValue) this.props.onSubmit({ value });
+    const cannotSubmit = (value === '') && !this.props.canClear;
+    const validValue = (cannotSubmit) ? previousValue : value;
+    this.setState({ value: validValue, editing: false });
+    if (value !== previousValue) this.props.onSubmit({ value: validValue });
```
ben-burlingham: nb Convention surround `if` with `{` `}`...can discuss if that's a thing or not
@chrisarevir

chrisarevir: Yeah, it may seem verbose, but it becomes pretty clear what it's encapsulating. I would vote for the surrounding braces, but NB for now

PR #344, opened 2017-09-09
Refactor Column Resizing

nickwmurphy:
Turns out that directly mutating the props is the best way to do this.

Code was going to become unmaintainable with the previous method, especially if columns were going to maintain their width when being reordered.


table/src/BodyRow.jsx
```diff
@@ -24,8 +23,8 @@ const BodyRow = ({
       key={`cell-${rowData[idKey]}-${def.key}`}
       className={cx(styles.tableCell)}
       style={{
-        flex: `1 0 ${columnWidths[index]}px`,
-        minWidth: `${columnWidths[index]}px`
+        flex: `1 0 ${def.width}px`,
+        minWidth: `${def.width}px`
```
ben-burlingham: ❤️

PR #348, opened 2017-09-11
Tabs: Analytics.

ben-burlingham:


tabs/sandbox/sandbox.jsx
```diff
@@ -1,29 +1,39 @@
...
-import Tabs from '../src';
+console.info('TABS_ANALYTICS enum:', TABS_ANALYTICS);
 
-const onChange = () => {
-  console.warn('Tab changed; external callback triggered.');
-};
+const onAnalytics = (obj) => { console.warn('Analytics: ', obj); };
+const onChange = () => { console.warn('Tab changed.'); };
 
+// TODO use `data` for consistency? Ben 170909
```
chrisarevir: :+1:

tabs/src/Tabs.jsx
```diff
@@ -46,15 +53,21 @@ class Tabs extends React.Component {
...
+  onAnalytics = (event) => {
+    this.props.onAnalytics({
+      event,
+      particle: TABS_ANALYTICS.NAME,
+      scope: this.props.scope
+    });
   }
 
-  setSelected({ k, v, index }) {
+  onChange = ({ k, v, index }) => () => {
```
chrisarevir: Don't need the second function call here!

ben-burlingham: Checked, existing code that sets up a closure with proper `k, v`

PR #359, opened 2017-09-11
Created Promisify shim

mboperator:
## Why
Shimming `promisify` because it's only available in Node v8.4.0+
Also did some linting.

cc @mcclayton @ben-burlingham


mcclayton:
Thanks @mboperator 🚀

mcclayton:
Note: You can remove step 0 from https://github.com/procore/particles#how-do-i-get-started

ben-burlingham:
Rather than requiring our users (who may not be engineers) to all debug and upgrade their node installations, is there a more backwards-compatible solution we could use?

Either, write vanilla promises, or use an existing NPM package, or...lots of options here that don't block people.

mboperator:
@ben-burlingham This is the backwards compatible solution, if users are using a Node version prior to 8.0 the script will utilize the shim.

ben-burlingham:
Still confused why we need all these updates to support vanilla functionality.

ben-burlingham:
OK reading thanks!

PR #360, opened 2017-09-11
Bulk test script

mboperator:
## Why
Once we get into using a CI tool, we're going to want to traverse through all the `particles` and run each of their test suites. This was the simplest way to do it, without making any assumptions about how each particle is being tested.


scripts/init.js
```diff
@@ -0,0 +1,34 @@
...
+  findJavascriptProjects,
+  prettyStringify,
+  writeFile
+} = require('./utils.js');
+
+/*
+  Helpers
+*/
+const installDependenciesForPath = (path) => {
+  const command = `cd ${path} && yarn`;
```
chrisarevir: Not all machines have yarn when they try and do things, even if it says in the README, should probably spit out a nice error if they don't at this point.

chrisarevir:
Otherwise, looks solid!

scripts/init.js
ben-burlingham: Yarn's tricky because it's so nice, but you're right, it's one more dep that we've been able to avoid thus far

scripts/test.js
```diff
@@ -0,0 +1,39 @@
...
+      exec(command)
+        .then(result => resolve({ path, output: result, testPassed: true }))
+        .catch(e => resolve({ path, output: e, testPassed: false }))
+  );
+};
+
+/*
+  Install subfolder dependencies
+*/
+const installSubDependencies = () =>
```
mcclayton: Since this is nearly identical code, can we share this between test and init?

scripts/test.js
```diff
@@ -0,0 +1,39 @@
...
+/*
+  Helpers
+*/
+const runTestCommand = (path) => {
+  const command = `cd ${path} && yarn test`;
+  console.info(`Testing ${path}`);
+  return new Promise(
+    resolve =>
+      exec(command)
+        .then(result => resolve({ path, output: result, testPassed: true }))
```
mcclayton: This is a really nice simple interface. 👍 
Would be kind nice, instead of simply logging to console `Testing bulk-editor`, `Installation complete! Check results in test.log`, etc. 
To simply log to console something like:
```
Testing all particles... 

Test Suite (bulk-editor) ✗
Test Suite (table) ✔
...
----------------------
Test Suites Passed: 1
Test Suites Failed: 1
```

scripts/init.js
mboperator: Ahh good point. I can switch all the scripts to use `npm` instead. Yarn is just habitual at this point haha.

scripts/test.js
mboperator: Nice I love that. I'm not quite sure how to do something like that via console logs (since the checkmark is dynamic) I'll look into it though. Also open to library suggestions

mcclayton: In your promise chain you can just have a post-process step that reads that file and prints the appropriate test result summary:
```
.then(results => postProcessTestResults('test.log', results))
```
If we do this, we might want a more descriptive name other than `test.log` such as `testResults.json`.

Or something to that general effect. Thoughts?

mboperator: Yeah, so I thought you proposed that one line is printed when the test starts, and then later modified when the test completes. IMO that'd be the slickest way to do it.
 
Just console logging after each test runs is much more straightforward so we'll go with that.

mcclayton: Ah yeah, no for something like that you'd need to use terminal buffers or cursor positioning. Maybe a good follow-up just for slick-ness 🚀 
Just logging is good for now 👍

scripts/init.js
chrisarevir: We can keep yarn, just gotta add the check that it's there ^^;

mboperator:
In an attempt to keep our PR tab clean, I'm going to close this temporarily until I revisit it later this week

PR #362, opened 2017-09-11
variable height for modal

thewillhuang:


modal/sandbox/sandbox.jsx
```diff
@@ -19,7 +19,7 @@ class Component extends React.Component {
           closeOnOverlayClicked
           show={this.state.show}
           closeModal={() => this.setState({ show: false })}
-          size='large'
+          size='medium'
```
ben-burlingham: This should be enumerated somehow.

Also, the 2.0 version of this modal should remove the need for sizing...other Particles would achieve this by overriding styles and inserting content in slots.

@Marswang92 what's the status on the Modal fixes?

modal/src/Modal.jsx
```diff
@@ -75,7 +76,7 @@ class Modal extends Component {
     );
 
     return show ? (
-      <div className={styles.container}>
+      <div className={styles.container} style={{ ...{ height: this.props.height } }}>
```
ben-burlingham: This should be achieved by overriding styles.

thewillhuang: how does overriding sytles work?

thewillhuang: actually, I think using the style composer isn't quite as good as using a prop that would just override. it's obvious what height would do. but not obvious what stylesheet would do.

ben-burlingham: Adding props to override individual styles doesn't scale that well.

The stylesheet composer allows you to override any part of the stylesheet, including height of any container.  

The upcoming docs will expose the base stylesheets for each particle, so the dev (or UX designer) can see what to override.  This allows us to work around `application.css` with its `!importants` and also to theme any component to fit any product :)

thewillhuang: okay, so perhaps this pr isn't needed

thewillhuang: great point

ben-burlingham:
@thewillhuang Mars' new modal https://github.com/procore/particles/pull/365

Issue #317, opened 2017-08-31
Modal 2.0 proposal


Marswang92:
1. Remove `beforeOpen`, `afterOpen`, `beforeClose`, `afterClose` props.
2. `size` props decide the modal's `max-height` and `width`.
3. Lock body scroll when modal is open.
4. Remove `disableHeader` prop.
5. Introduce `CSSTransition` for animation.

ben-burlingham:
Sounds good just a couple requests:

- `size` is solved with stylesheet
- Let's use the `slot` pattern from other components: `slotHead`, `slotBody`, `slotFooter`

ben-burlingham:
https://github.com/procore/particles/pull/365

PR #365, opened 2017-09-12
Modal 2.0

Marswang92:
# Change Log
1. Use `onOpen`, `onClose` props instead of  `beforeOpen`, `afterOpen`, `beforeClose`, `afterClose`.
2. `size` prop removed. Instead user can set modal size through `stylesheets` prop.
3. Remove `disableHeader` prop.
4. Unit test introduced.

## related issue
https://github.com/procore/particles/issues/317


modal/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["latest", "react", "stage-0"]
```
ben-burlingham: Is this file still necessary?

modal/sandbox/sandbox.scss
```diff
@@ -1,3 +1,9 @@
+// UX Approved Color Pallette
+$black1: #343739;
+$gray3: #c3c2c2;
+$gray4: #e7e7e7;
+$white1: #fff;
```
ben-burlingham: @nickwmurphy do these align with UX particles palette

modal/sandbox/sandbox.jsx
```diff
@@ -21,9 +21,10 @@ class Component extends React.Component {
           closeModal={() => this.setState({ show: false })}
           size='large'
```
ben-burlingham: `size` was removed

modal/sandbox/sandbox.jsx
```diff
@@ -21,9 +21,10 @@ class Component extends React.Component {
...
           slotHeader={'Header'}
-          slotFooter={(<div className={sandboxStyles.footer}>Footer</div>)}
-          onShow={show => this.setState({ show })}
-          onHide={show => this.setState({ show })}
+          slotFooter={(<div>Footer</div>)}
+          stylesheets={[sandboxStyles]}
+          onOpen={() => { console.log('open'); }}
+          onClose={() => { console.log('close'); }}
```
ben-burlingham: nb The other sandboxes do these declarations differently, I think for consistency we should follow them but no big deal.  Maybe we should talk about what standards we want on the sandboxes.

modal/src/Modal.jsx
```diff
@@ -9,27 +10,19 @@ const isClosing = (s1, s2) => s1.show && !s2.show;
...
   componentDidUpdate(prevProps) {
-    if (isOpening(prevProps, this.props) && this.props.afterOpen) {
-      this.props.afterOpen();
+    if (isOpening(prevProps, this.props) && this.props.onOpen) {
+      this.props.onOpen();
     }
 
-    if (isClosing(prevProps, this.props) && this.props.afterClose) {
-      this.props.afterClose();
+    if (isClosing(prevProps, this.props) && this.props.onClose) {
```
ben-burlingham: Shouldn't need `&& this.props.onClose` if default props have declared it to `() => {}`

modal/src/Modal.jsx
```diff
@@ -43,67 +36,55 @@ class Modal extends Component {
...
-        [styles['modal-small']]: size === 'small',
-        [styles['modal-medium']]: size === 'medium',
-        [styles['modal-large']]: size === 'large'
-      }
-    );
     const closeClassName = cx(
-      styles.close, {
-        [styles['close-black']]: disableHeader
+      this.styles.close, {
+        [this.styles['close-black']]: !slotHeader
```
ben-burlingham: nb This feels funny.  Seems like there should be one close style.  I see what you're going for here though...maybe we can iterate in the future.

modal/src/Modal.jsx
```diff
@@ -43,67 +36,55 @@ class Modal extends Component {
...
-          )}
-          <div className={styles.content}>
-            { children }
+      <div className={this.styles.container}>
+        <div className={this.styles['modal-window']}>
+          <div className={this.styles['modal-header-container']}>
+            <div className={closeClassName}>
+              <i
+                className='fa fa-close'
+                onClick={this.props.closeModal}
```
ben-burlingham: Shouldn't this call `onClose` from the props?

modal/src/Modal.jsx
```diff
@@ -120,35 +101,25 @@ Modal.propTypes = {
...
-  afterClose: PropTypes.func,
-  onCloseClicked: PropTypes.func,
+  onOpen: PropTypes.func,
+  onClose: PropTypes.func,
   closeModal: PropTypes.func
 };
 
 Modal.defaultProps = {
   stylesheets: [],
   children: <div />,
```
ben-burlingham: IMO `slotBody` would be more appropriate since follows other props

modal/sandbox/sandbox.jsx
ben-burlingham: ^ @mboperator

modal/.babelrc
mcclayton: With Jest, yes.
https://github.com/facebook/jest#using-babel
There may be a way around it, but a `.babelrc` file is the recommended approach.

modal/test/Modal.test.jsx
```diff
@@ -0,0 +1,62 @@
...
+  const mergedProps = Object.assign({}, props, newProps);
+  return mount(
+    <Modal {...mergedProps} />
+  );
+};
+
+let showWrapper;
+let hideWrapper;
+
+beforeEach(() => {
```
mcclayton: `nb`
I don't believe this needs to be in a `beforeEach`.
You can simply create two consts:
```
const showWrapper = renderModal({ show: true });
const hideWrapper = renderModal({ show: false });
```

modal/src/Modal.jsx
Marswang92: yes mb

Marswang92: Cuz previously there's a callback specifically for user click on the close button

modal/src/Modal.jsx
Marswang92: I think proper way to do it is just let user specify the style of the close button. Anyway we can figure out according to the people's feedback on using this.

Issue #368, opened 2017-09-12
App crashes on tableData = [] as of version 7.0.0


scottwalstead:
Sending in tableData = [] crashes the app as of version 7.0.0

One solution would be to convert tableData = [] to tableData = [{data: []}]

Another solution would be to include messaging that tells particles consumers not to send in tableData = []

PR #371, opened 2017-09-13
Table: tableData fallback for passing in an empty array

scottwalstead:
The update for v7.0.0 deprecated consumer ability to pass in tableData as an empty array. Our two options as temporary fixes are
1) create a fallback that converts the empty array to the data structure we are expecting 
2) we can message consumers telling them to pass in the data structure we are expecting

Talking to Ben, it sounds like there might be changes coming to the data structures so I recommend going with the fallback solution for the time being to minimize the impact on clients having to update their code.

Fixes https://github.com/procore/particles/issues/368


ben-burlingham:
Just FYI, the changes I was referring to were for the possible updates supporting N-Tier grouping.  Not sure when this will happen.

PR #375, opened 2017-09-13
Storyshots initial implementation

mboperator:
## Changes
This implements `Storyshots` for snapshot testing our React components.

## Example Run
screen shot 2017-09-13 at 3 34 34 pm

## Error Diffs
screen shot 2017-09-13 at 11 58 52 am


mcclayton:
From a code/setup perspective this looks good to me!

We already discussed my main concerns, but want to post them here for visibility.
1. We need to be clear to developers that this is **not** a drop in replacement for unit tests.
2. The danger is that people just update snapshots without actually being able, or taking the time, to understand the cause/effect changes. It becomes an ignorable diff when used improperly. I've seen it happen before where teams just add `yarn test -- -u` to their check-in procedure without actually diving into what effect their changes are causing.

Just my `2¢`. I definitely think this will add value as long as we don't abuse it 🚀

mboperator:
These are fantastic to bring up and I don't want them to get lost in the shuffle.
As per our convo offline, I'll update the README.md to add a blurb about Jest snapshot testing and resisting the `test -- - u` temptation 👍

mcclayton:
Discussed in person - Would be good to perhaps add instructions on how to run/update the snapshots
Can be added to: https://github.com/procore/particles#yarn-test

As well as including a summary of the disclaimers I listed above to reduce abuse.

Optional, perhaps add a link to our T[esting Guide](https://procoretech.atlassian.net/wiki/spaces/DEV/pages/169844404/UI+Unit+Testing) as well.

mboperator:
Learning to document in person convo's from the best @mcclayton 😎

README.md
```diff
@@ -50,6 +50,19 @@ In the tool which is importing `FOO`, use `yarn link FOO` to take advantage of t
...
+All Storybook documented `particles` get a thin layer of snapshot testing provided by `Storyshots` and jest.
+
+To update the snapshot to support your `new Particle` or `sandbox change` run:
+```
+yarn test -- -u
+```
+
+** Be Warned **
+1. This is **not** a drop in replacement for unit tests.
+2. This is **not** a drop in replacement for unit tests!
```
mcclayton: 😍

michaelghinrichs:
very cool!

PR #381, opened 2017-09-14
Table, Filters, Spinner: Scoping fixes.

ben-burlingham:


mcclayton:
With https://github.com/procore/particles/pull/375 being merged, most all changes will break our snapshot tests (not a bad thing). The snapshots are already red and out of date. Do we have any plans to get particles onto CircleCI and so we can start hooking up tests to our build process?

ben-burlingham:
@mcclayton The current strat was to hook test stuff into the prod build script so it's performed locally and can be fixed immediately.  We can bring this up in a chapter meeting though?

PR #383, opened 2017-09-14
Enable Mock CSS Modules for tests.

mcclayton:
Currently, we use the [identity-obj-proxy](https://www.npmjs.com/package/identity-obj-proxy) in order to mock our css modules for testing:
https://facebook.github.io/jest/docs/en/webpack.html#mocking-css-modules

This normally would work straight out of the box, but  our stylesheetCompose tramples over the proxy object.

This change will check to see if the base style being passed in is the `identity-obj-proxy`, and will return it without composition in the composer. This will enable us to actually see CSS classNames in our tests and be able to assert on them as well as `find` them using enzyme without having to have conditional logic in our components like this: https://github.com/procore/particles/blob/master/modal/src/Modal.jsx#L13

Using the super sick storyshots feature @mboperator added, we can get a glimpse into how we'll now be able to view classNames in our tests.

screen shot 2017-09-14 at 10 53 54 am


mboperator:
This is epic!

PR #386, opened 2017-09-14
Move sorting arrow icon next to HeadCell label

scottwalstead:
Move sorting arrow icon next to HeadCell label to prevent far right icon alignment bug introduced in 6.1.0

@TODO add GitHub issue with screenshot
@TODO get feedback from Nick Murphy


ben-burlingham:
We should add sorting functionality to the sandbox so we can see this stuff - not sure why it's not there.  I'll add to my list.

PR #388, opened 2017-09-15
Dialog 1.0

thewillhuang:
new api


modal/src/index.js
```diff
@@ -1,3 +1,50 @@
...
+Modal.defaultProps = {
+  visible: false,
+  children: [],
+  style: {},
+  className: '',
+  onClickScrim: () => { console.log('clicked scrim'); }
+};
+
+
+export const Footer = ({ style, className, children }) => (
```
mboperator: How do you feel about breaking these out?
Also it'd be awesome if we had `propTypes` for these.

modal/sandbox/sandbox.scss
```diff
@@ -17,16 +17,16 @@ $white1: #fff;
...
-  box-shadow: 0 0 12px $black1;
-  display: flex;
-  flex-direction: column;
-  left: 100px;
-  min-width: 500px;
-  min-height: 200px;
-  overflow: auto;
-  padding: 0;
-  z-index: 101;
+  // background-color: $white1;
```
mboperator: Should we delete this?

thewillhuang:
im game lets do it

ben-burlingham:
What's the advantage here?

thewillhuang:
control over styles

thewillhuang:
The main point is not the extra components, its only there to get you easy standards. Mainly, this rewrite allows for _flexibility_. You may use those extras if you just want a simple way to get the old styles, but they are not necessary

ben-burlingham:
For styles, there's patterns in place everywhere already to do that, namely the stylesheet composer. 

Passing in a class name opens the CSS to the global namespace and will cause bugs down the line when things are changed in one place, and affect another (like current Procore).  

By using the existing pattern of extending the scoped CSS, not only can we target any piece of the layout, but the CSS remains scoped.  

For flexibility, I'm not sure what you are referring to?  The current design has slots for the head, body, and footer.  You render what you need, with styles scoped from your tool, and pass it in.

Do you mind pulling this out to a different particle, since it doesn't follow any existing conventions and is an alternative design?

mboperator:
So, we looked at the stylesheet composer and passing in a `className` does the exact same thing. IE: `overrideStyles.modal-header` would still resolve to `.modal-header-mdq3`. 

The upside is the developer doesn't need to know exactly which styles they need to override. If they want to override the header, well they just send either inline styles or a classname to `<Modal.Header>`. In my opinion this is `simpler` and more `explicit` because if you want to style the header, you style the header instead of sending in a `map` full of styles with certain classes you need to know to overwrite.

Additionally this pattern is more declarative which matches the React ecosystem, and as @thewillhuang said it's composable which matches the values of being `explicit`.

Our main goal with this refactor was to revisit the API since the changes necessary to support the new styles would require us to change the markup anyway. I'd like to maintain only one modal, since it's a core component (like Buttons, and Selects). We can fork it, but I think developer experience would suffer and cause further fragmentation.

ben-burlingham:
OK, thanks for the clarification.

Can we compromise by adding support just for additional classNames, but leaving the original markup?

nickwmurphy:
I have been developing in Particles for a few months now, and I am struggling to see how this change makes a difference in a dev's workflow. Also, it feels like this new convention could snow ball into breaking changes across all particles. What about the current convention is not working?

modal/src/Modal.scss
```diff
@@ -33,21 +40,15 @@ $white1: #fff;
 .content {
   flex-grow: 1;
   width: 100%;
+  padding: 15px 15px;
```
nickwmurphy: `padding: 15px` is cleaner

modal/src/Modal.scss
```diff
@@ -33,21 +40,15 @@ $white1: #fff;
...
-
-.header {
-  background-color: $black1;
   border-top-left-radius: 2px;
   border-top-right-radius: 2px;
-  color: white;
-  display: flex;
-  padding: 15px 20px;
-  position: relative;
+  padding: 15px 15px;
```
nickwmurphy: `padding: 15px` is cleaner

modal/src/Modal.scss
```diff
@@ -4,6 +4,17 @@ $gray3: #c3c2c2;
 $gray4: #e7e7e7;
 $white1: #fff;
 
+.header {
+  background-color: $black1;
+  border-top-left-radius: 2px;
+  border-top-right-radius: 2px;
+  color: white;
```
nickwmurphy: `border-radius: 2px 0` is cleaner

modal/src/Modal.scss
```diff
@@ -4,6 +4,17 @@ $gray3: #c3c2c2;
 $gray4: #e7e7e7;
 $white1: #fff;
 
+.header {
+  background-color: $black1;
+  border-top-left-radius: 2px;
+  border-top-right-radius: 2px;
```
nickwmurphy: `border-radius: 2px` 0 is cleaner

modal/src/Modal.scss
```diff
@@ -6,11 +6,10 @@ $white1: #fff;
 
 .header {
   background-color: $black1;
-  border-top-left-radius: 2px;
-  border-top-right-radius: 2px;
+  border-radius: 2px 2px 0 0;
```
brock-noah: I disagree with [this](https://github.com/procore/particles/pull/388#discussion_r139254606) suggestion. You should only apply properties that you actually want. Do you really want the bottom radii `0`? In this case probably yes, but the default value is already `0`! You are potentially messing up some CSS defined elsewhere. Never apply CSS that isn't necessary.

`margin: 0 auto` drives me crazy. With single class specificity it is not too bad, but imagine if this was nested. Now you have higher specificity and from the "cleanliness" more code needs to be written elsewhere.

modal/src/Modal.scss
```diff
@@ -4,6 +4,16 @@ $gray3: #c3c2c2;
 $gray4: #e7e7e7;
 $white1: #fff;
 
+.header {
+  background-color: $black1;
+  border-radius: 2px 2px 0 0;
```
brock-noah: I disagree with [this](https://github.com/procore/particles/pull/388#discussion_r139254606) suggestion. You should only apply properties that you actually want. Do you really want the bottom radii `0`? In this case probably yes, but the default value is already `0`! You are potentially messing up some CSS defined elsewhere. Never apply CSS that isn't necessary.

`margin: 0 auto` drives me crazy. With single class specificity it is not too bad, but imagine if this was nested. Now you have higher specificity and from the "cleanliness" more code needs to be written elsewhere.

brock-noah:
Why I like
```
// A
<Modal>
  <Modal.Header />
  <Modal.Body />
</Modal
```
over
```
// B
<Modal
  headerSlot={}
  bodySlot={}
/>
```
is because if you have an implementation like A, you can create a component like B, and keep greater future flexibility.

```
<ModalWithSlots
  header={}
  body={}
/>

const ModalWithSlots = ({ header, body }) => (
  <Modal>
    {header && <Modal.Header>{header}<Modal.Header />}
    {body && <Modal.Body>{body}<Modal.Body />}
  </Modal>
)
```

Now imagine the `Footer`, which is more like `Modal.ActionBar`, is moved beneath the header. In version B a new prop needs to be introduced to the component, while in version A you just move it to where you want.

Modals are generally pretty consistent in design, where as Cards have many possible combinations of content. I really like this [example](http://react-toolbox.com/#/components/card) showing how `Card` defines types and you are free to move those types around to declare what you want. If a card was created with slots, you'd run into props like `headerAboveImage`, `headerBelowImage`, `imageBelowHeaderButAboveText`

modal/test/Modal.test.jsx
```diff
@@ -1,56 +1,64 @@
 import React from 'react';
-import { mount } from 'enzyme';
-import sinon from 'sinon';
-import Modal from '../src';
-import baseStyles from '../src/Modal.scss';
+import renderer from 'react-test-renderer';
```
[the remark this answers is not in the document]
thewillhuang: really, i'm waiting on https://github.com/facebook/react/pull/7409 I believe currently it's too brittle but gets the job done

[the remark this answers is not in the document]
mcclayton: Was there any particular reason you chose not to use enzyme? They have a pretty great selector API.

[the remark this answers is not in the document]
thewillhuang: Snapshot testing is the future. DX is first rate and a devs time is spent on the most important which is designing a functional API that is easily testable.  Instead of trying to waste time mocking with sinon, you can just execute the function and see the real result in a snapshot.

The only thing it does not do is making sure that the CSS renders correctly in all browsers, but enzyme does not do that either so I'm not seeing much benefit.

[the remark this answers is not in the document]
mcclayton: Let's sync up on Tuesday :)

ben-burlingham:
Those are some great examples Brock, thanks for taking the time to flesh them out.

> Now imagine the Footer, which is more like Modal.ActionBar, is moved above the body, maybe even no header (like a confirm dialog with buttons at the top). In version B a new prop needs to be introduced to the component to determine where those buttons should be placed, while in version A you just move it to where you want.  

First off, footers should never be above the body, by definition.  But if you did want to put feet above torsos, then just pass in your markup to the header slot.  No need for extra props or anything, just pass in your confirm button container to the header slot.

> Modals are generally pretty consistent in design

Yep.  Header, body, footer, optional close button.  All containers.  Markup and styling will be dependent on application.

chrisarevir:
Late to the game, but wouldn't it be easier/more extensible if the modal worked like this?

```
<Modal i18n={title: 'I am a title!' submit: 'submitText, ...}>
    <div>
        Body things here
        <span>Example Text here</span>
    </div>
</Modal>
```

and then if you pass a headerRenderer it uses that for the "Header" portion of a footerRenderer for the "Footer" portion it uses those instead? Seems to be a pretty nice standard in other places, and allows for greater flexibility without reinventing the wheel. See [here](https://jquense.github.io/react-widgets/api/Multiselect/#tagComponent) where it passes in `tagComponent` as the render function for a tag.

I only mention this because I spent a lot of time trying to do just this for the bulk editor, but it came back to this in the end because it was actually simpler and easier to implement/work with.

mboperator:
@chrisarevir In my opinion, the example API that you've linked isn't more extensible, because it makes assumptions about a `title` and `submit` button at the parent level. Everything is locked in when using a slot paradigm.

IE the equivalent with a more declarative API:
```jsx
<Modal>
    <Header><h1>{i18n.title}</h1></Header>
    <Body>
      Body things here
    </Body>
    <Footer>
      <Button>{i18n.submit}</Button>
    </Footer>
</Modal>
```
This API is _declarative_ in that it's very clear where the `i18n` variables are going. There's no magic/no top level component controlling where things go. The developer has control over where things go.

It could be argued that this requires more boilerplate, but because of how easily composable everything is, an abstraction is easy:

```jsx
const SubmitFooter = ({ onSubmit, label }) => (
  <Footer><Button onClick={onSubmit}>{label}</Button></Footer>
)

const CoolModal = () => (
<Modal>
    <Header><h1>{i18n.title}</h1></Header>
    <Body>
      Body things here
    </Body>
    <SubmitFooter onSubmit={handleSubmit} label={i18n.submit} />
</Modal>
```
)

ben-burlingham:
Why can't you make a collection of `<Footer>` components in the application (Procore, marketplace, whatever), compose them as you wish, and pass them into the footer slot?  One modal footer ~may not~ will not be the same as the next.  

Please separate this out into a new particle.

thewillhuang:
okay, will make this into a new particle, thanks for the input guys

ben-burlingham:
Thanks Will, we'll start using it!

brock-noah:
HTML is more familiar to developers than self closing tags, passing JSX like objects, and writing more CSS when you repurpose a slot. I think we should make components look like HTML because it is what developers are accustomed to.

thewillhuang:
@ben-burlingham okay, thanks man

dialog/sandbox/sandbox.scss
```diff
@@ -0,0 +1,17 @@
+// UX Approved Color Pallette
+$black1: #343739;
+$gray3: #c3c2c2;
+$gray4: #e7e7e7;
+$white1: #fff;
+
```
nickwmurphy: Recommend `@import '../../shared/variables';` and using the ones available in that file.

dialog/src/Dialog.scss
```diff
@@ -0,0 +1,76 @@
+// UX Approved Color Pallette
+$black1: #343739;
+$gray3: #c3c2c2;
+$gray4: #e7e7e7;
+$white1: #fff;
+
```
nickwmurphy: Recommend `@import '../../shared/variables';` and using the colors available in that file.

dialog/test/Modal.test.jsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import Modal, { Header, Body, Footer } from '../src';
+
+class ModalWithState extends React.Component {
```
mcclayton: Will sync more with you on these tests later, but the filename and wording should all be changed to say `Dialog` instead of `Modal`

thewillhuang: woops missed it! thanks!

mcclayton: Synced in person with @thewillhuang. Decided to remove tests from this PR for now and we'll pair on writing the unit tests in a separate PR.
Some summary discussion about snapshots:

**The summary point I'd like to make is that Jest snapshots are **NOT** a drop-in replacement for unit tests.
For particles specifically, I'd like to have storyshots be the only place we use snapshots and that all the test directories be "actual" unit tests.**
Reasons being: 
1. Cannot practice TDD with snapshots
2. Lose the advantage of finding bugs through writing tests as your snapshot would just encapsulate buggy code but still be green.
3. Snapshots tend to fail when code is changed, not when behavior changes.
4. Snapshots do not do a good job of asserting behavior at the unit level. This is where sinon.js spies have much better behavioraral and API assertion capabilities.
	**->** For example, with these tests assertions do not assert the behavior they say they do. Tests should fail when they no longer do what their `describe() it()` statements say they should do. These tests fail even though the behavior being assertet did not change. i.e. 
	'modal' 'overlay onClickScrim' '✕ hides when triggered onClickScrim (5ms)' fails when we add any class to the header.
5. When you have a ton of snapshots, they will break *all* the time and then the updating of snapshots becomes abused and it's difficult to make sense of what (if anything) actually broke.

That being said, I think snapshots can provide a lot of value when used correctly and I think our storyshots is a great use for them.

PR #395, opened 2017-09-21
Allow only yarn as package manager, fixes #394

### LINKED ISSUE #394: Only allow usage of Yarn
- Add `package-lock.json` to `.gitignore`
- Document that `Yarn` is the approved package manager.
- Trash any `package-lock.json` related files

ryanyogan:
### Summary
- Added `package-lock.json` to the `.gitignore` file to force the usage of one package manager, yarn.
- Addresses issue #394 

- Make sure engineers know that Yarn is the official package manager.
- If a user utilized npm the node_modules and package.json will be updated, this will cause issues with the `yarn.lock` file, we could shim or override any npm calls within this directory to alias to `yarn` if desired?

### Testing
- Attempt to use `npm i package-name` the `package-lock.json` file should exist however should be ignored by git.
- Test that adding the same package via `yarn add package-name` does modify the `yarn.lock` file.


README.md
```diff
@@ -50,6 +50,11 @@ In the tool which is importing `FOO`, use `yarn link FOO` to take advantage of t
 Starts the appropriate test runner for `FOO`'s test suite and runs the corresponding test file(s) from within the `/test` directory.
 Also generates a code coverage report for `FOO` component and prints it to the console.
 
+
+#### Notes
+
+`yarn` is the officially supported Node Modules package manager, if you use NPM locally your `package-lock.json` will be ignored and the `yarn.lock` file will not be updated properlly.  Please only use `yarn install` and `yarn add package-name` to avoid deployment issues.
```
mcclayton: `properlly` -> `properly`

ben-burlingham:
We've managed to avoid using Yarn so far, and global requirements in general are purposely few and far between.

The discussion on guild frontend was for Wrench - the build pipeline for individual particles is up to the implementer.

chrisarevir:
I will agree with @ben-burlingham here. Since each individual particle is discrete and doesn't have an automated build step, nor should it, the need to choose one or the other is just that, a choice.

As it stands, most everyone here uses yarn anyways, but it is not a requirement. If we ever wanted to enforce it, it should be opened up to the chapter as a discussion topic / issue and then make it concrete in the README which to use after that.

PR #397, opened 2017-09-22
allow right controls in filter bar

evinism:
### Summary
Adds in a possible right control to the filter bar in list-layout. In my screenshot, toggler between the a thumbnail and list view.

#### Before
screen shot 2017-09-22 at 9 29 40 am

#### After
screen shot 2017-09-22 at 9 24 42 am

The toggler itself should probs be a particle, i'll do that too.


list-partial/src/ListPartial.jsx
```diff
@@ -24,6 +25,12 @@ const ListPartial = ({
     </div>
   );
 
+  const rightControl = rightControlComponent && (
+    <div>
```
ben-burlingham: nb Perhaps consider putting a class on this wrapper so devs can target and override the `slotRight`?

ben-burlingham:
Also, have you considered placing this in the filter component instead?  The current thinking of the list partial is to sort of "stack together" particles for a layout, rather than customize the layout. What do you guys think with respect to your tool?

evinism:
@nickwmurphy can do, thanks!

@ben-burlingham i chose this path because
1: it's not a breaking change
2: given how search and filter are semantically separated from each other, it seemed silly to stick what in my use case is a view switch into the same slot as a filter
3: the same thing is done for the left slot in this component

that relatively sensible, or would you still prefer preventing a code change by sticking it in the filter param over this?

ben-burlingham:
Nope, makes sense! Thanks!

evinism:
whoops, forgot to rebuild index, don't merge until i've done that.

PR #399, opened 2017-09-23
Table - Ability to disable column auto resize

scottwalstead:
### Summary
Our team had an issue where the edit column took up half of the overall table size when we wanted it to only take up about 10% of it. By making this change, we are able to guarantee that the column size isn't larger than intended when we render the table.

#### Before
screen shot 2017-09-22 at 11 35 45 am

#### After
screen shot 2017-09-22 at 6 06 12 pm

Altered the sandbox column definitions to clean them up a little bit and make it easier to test the autoResizable flag.

### Testing

- [ ] If columns don't fill the width of the table (no side scrolling), they auto resize to fill the table width by default.
- [ ] Any columns that have autoResizable set to false should be the width that they are set to in the column definition
- [ ] If only 2-3 small columns (less than 200px) are prevent and they all have autoResizable: false, then the last column will be the auto resize column and will fill the width of the table


table/src/Table.jsx
```diff
@@ -338,6 +338,17 @@ export default class extends React.Component {
 
     const tableData = this.props.tableData.length > 0 ? this.props.tableData : [{ data: [] }];
 
+    // Add default prop autoResizable: true to column definitions if undefined
+    columnDefinitions.forEach((def) => {
+      def.autoResizable = (typeof def.autoResizable !== 'undefined') ? def.autoResizable : true;
+    });
```
ben-burlingham: This may not be necessary?  The checks above are:

`!def.autoResizable ? ...` which should trigger on `undefined`.

Also, populating this each time in the render method so will affect perf - there's probably a better place for it?

table/src/Table.jsx
```diff
@@ -338,6 +338,17 @@ export default class extends React.Component {
...
+
+    // If all visible columns are autoResizable: false, make at least one autoResizable: true
+    // autoResizable refers to the table resizing column widths to fill the entire table width if necessary
+    if (columnDefinitions.every((def) => !def.hidden && !def.autoResizable)) {
+      columnDefinitions[columnDefinitions.length - 1].autoResizable = true;
+    }
```
ben-burlingham: I can see a use case here where someone might want an explicit width on every column and intentionally set `autoResizable=false` on all of them.  Is it wise to second-guess this case?  @nickwmurphy @yoyis3000 @piercefrawley @evandapolarbear have used tables pretty heavily, any opinions here?

Also , same as above - this probably doesn't have to happen on every render?

table/src/Table.jsx
scottwalstead: Since we want it to be true by default, I believe we do need it. For users to disable the resize, they need to set it to false.

table/src/Table.jsx
scottwalstead: So if the width of the columns add up to greater than the overall width, then this boolean won't have an effect on the column widths and they will render with the user specified size. It only has an impact on if the column widths are not wide enough to fill the table size, then one column has to become slightly larger to fill the table.

nickwmurphy: We will always need at least one column to be `autoResizable=true`. I think this code is necessary to ensure that. However, maybe we should `console.warn` if it is used? Also, I agree with @ben-burlingham that maybe we can find a place where this check isn't made for every render, but I am not sure where that would be...

scottwalstead: I still think there is a valid use case for this that wouldn't warrant a console warning because of TableConfig and hiding columns.

Example
Start with 3 columns `hidden: false`, first 2 have `autoResizable: false`
Config allows user to hide the third column and now the user sees only two columns
The two columns shown happen to be `autoResizable: false` but one needs to be set to `autoResizable: true`

In that scenario, I see that as a valid use case and therefore believe the warning isn't needed / doesn't help. 

The scenario it does help is a bad config in which every column (regardless of hidden true or false) is `autoResizable: false` and I think that deserves a warning message. Thoughts?

scottwalstead: @ben-burlingham @nickwmurphy in terms of placement, we probably shouldn't alter the props object. What I'm doing here probably isn't best practice. My thought is move this logic to the constructor and add columnDefinitions to the state object.

scottwalstead: Refactored the code out of render for validation of the columnDefinitions. We can add future validation to that function as well.

nickwmurphy:
👍 Looks great!

Issue #401, opened 2017-09-25
Node Security Module: `marked`


ben-burlingham:
[The module for interpreting markdown](https://www.npmjs.com/package/marked), `marked`, has been flagged by NSM for a potential DOS offender:

> The marked module is vulnerable to a regular expression denial of service. Based on the information published in the public issue, 1k characters can block for around 6 seconds.

The `marked` module is very widely used (35k daily downloads), and a third party package pulls it in as a dependency, so I'm not sure how to proceed here - whitelist?

ben-burlingham:
Resolution: 
- Remove `markdown-loader`. 
- Move `webpack-notifier` to dev dependency (thereby removing it from prod build)

Thanks @scottwalstead for sorting this out. :clinking_glasses:

PR #402, opened 2017-09-25
Proposal: NSM - removing `markdown-loader`

ben-burlingham:
### Summary
Node Security Module (NSM) has flagged the `marked` module as causing possible DOS on long strings.

The module is used in `markdown-loader` and `webpack-notifier`.  I'm removing `markdown-loader` from the Particles root. It looks like it's not used anywhere else.

Webpack notifier is a bit more tricky, I'll open a separate PR for that.

### Related Issues
Relates to: https://github.com/procore/particles/issues/401


scottwalstead:
If it's not being used, we should remove it. Approving.

PR #409, opened 2017-09-26
Adds props to keep track of the currently focused dropdown

justinmwatts:
### Summary
Adds props to allow us to keep track of the currently focused dropdown. This then allows us to support state re-renders when using a redux state management paradigm.

Add new props to the sandbox for documentation.

### Testing
Test a user is able to select and remove filters in the filters sandbox.


ben-burlingham:
@justinmwatts we'll be out for a bit, but @nickwmurphy is a Filters team member so he can answer questions or publish.  See you next week!

PR #411, opened 2017-09-27
Select Shift + Tab

nickwmurphy:
### Summary

1. Fixes a bug that was blocking the user from using `shift + tab` to navigate backwards on a page.

2. Bumps to React v15.6.2.

3. Fixes up sandbox to better show how the selects interact with one another.


select/src/Select.jsx
```diff
@@ -189,9 +190,31 @@ export default class Select extends React.Component {
...
+      this.setState({ currentIndex: -1, expanded, value: '' });
+    }
+  }
+
+  onMouseDown = () => {
+    if (document.activeElement === this.title || document.activeElement === this.caret) {
+      this.setState({ hasFocus: true });
+    }
+  }
+
```
nickwmurphy: `onMouseDown` fires before `onFocus`, giving a chance to let `onClick` to know if focus is already present.

select/src/Select.jsx
```diff
@@ -189,9 +190,31 @@ export default class Select extends React.Component {
...
       this.props.onFocus && this.props.onFocus(evt);
-      if (this.searchInput) this.searchInput.focus();
-      this.setState({ currentIndex: -1, expanded: !this.state.expanded, value: '' });
+      this.setState({ currentIndex: -1, expanded, value: '' });
+    }
+  }
+
+  onMouseDown = () => {
+    if (document.activeElement === this.title || document.activeElement === this.caret) {
+      this.setState({ hasFocus: true });
```
nickwmurphy: `onMouseDown` fires before `onFocus`, giving a chance to let `onClick` to know if focus is already present.

scottwalstead:
Not too familiar with this code but changes look good. 

At some point would be nice to discuss adding more comments to the code and how everyone feels about it. Little things like knowing onMouseDown fires before onFocus seems like it'd be useful as a comment in the code and not just the CR.

PR #413, opened 2017-09-28
Add onChange and onClear props

scottwalstead:
### Summary
Provides props to call for onChange and onClear

By doing this, we are able to update the parent state search value when it changes in the SearchBar

We needed to do this otherwise our stale initial search string would be passed to componentWillReceiveProps and wipe out our current search string in SearchBar state anytime our parent component did a render

### Testing
- [ ] Make sure onClear also calls this.props.onClear and onChange calls this.props.onChange


scottwalstead:
@mcclayton Yeah. I can write them. I assume I run the test suite with a command like `yarn test`

search-bar/src/SearchBar.jsx
```diff
@@ -48,13 +52,14 @@ export default class SearchBar extends React.Component {
...
+    this.setState({value: e.target.value});
+    this.props.onChange(e);
   }
 
   onClear = () => {
-    const value = '';
-    this.setState({ value });
-    this.props.onSubmit(value);
+    this.setState({value: ''});
+    this.props.onSubmit('');
```
mcclayton: General question for the group since I haven't followed this component's development as well as I should have - but why do we call `onSubmit` here? Do we really want to fire that, especially now that we have a specific prop function `onClear` to signify a clearing of text?

mcclayton:
@scottwalstead Thank you so much! Yep, you can simply run `yarn test` from within the search-bar component directory. There is already a test file `SearchBar.test.jsx` you can modify.

search-bar/src/SearchBar.jsx
scottwalstead: I think to ensure that we are showing results for an empty search. Otherwise the AJAX call would not be made.

We could force users to transition to using onClear but I like the idea of taking care of the onSubmit for them on form clearing.

mcclayton: My only qualm is that that can be achieved by the parent of this search-bar component via making use of this `onClear` prop. Right? Whereas the converse locks the parent in to this implementation. The other gives more freedom to the parent component and makes the child dumber. Maybe I'm missing something though as I've not seen many of the implementors of this component so that's just my `2¢`

scottwalstead: Yeah. I agree it does lock them in but it also protects them from the corner case that they may not think of making an AJAX call on clear. We could leave it until we get a complaint that this implementation doesn't work for someone. Also, if we change it now, we'd have to do a major version bump and have everyone consume onClear. 

For this PR, I only wanted to make it optional to pass in onClear since that seemed like a slightly bigger discussion than simply giving the parent access to this action handler.

scottwalstead:
@nickwmurphy Sounds good. Ideally we could add a command line lint task to make it IDE agnostic and git hooks could allow us to run linting before committing or pushing code.

search-bar/src/SearchBar.jsx
mcclayton: Yep, just wanted to raise the discussion :) I'm fine with leaving this as is and tackling a major bump when/if it becomes requested.

scottwalstead: Sounds good. We can add it as an issue for discussion on GitHub if we want to discuss it further or like you said just wait until if/when it gets requested.

scottwalstead:
@mcclayton Added unit tests for onClear and onChange

search-bar/test/SearchBar.test.jsx
```diff
@@ -59,6 +59,16 @@ describe('SearchBar', () => {
...
     });
 
+    it('should call appropriate onChange handler on change', () => {
+      const onChangeSpy = sinon.spy();
+      const mockChangeEvent = {
+        target: { value: 'foo' }
+      };
+      const wrapperWithSpy = shallowRenderSearchBar({ onChange: onChangeSpy });
+      wrapperWithSpy.find('input').simulate('change', mockChangeEvent);
+      expect(onChangeSpy.called).toEqual(true);
```
mcclayton: This can be even more specific with `expect(onChangeSpy.calledWith(mockChangeEvent)).toEqual(true);`

scottwalstead: will update. nice catch.

PR #414, opened 2017-09-28
Allow User's onBlur and onEnter functions to work after User Types Their Input

idannyou:
### Summary
Current Date Select won't allow users to register the date value after typing the Date and pressing Enter or clicking out of the Date Input Field.

Users should be able to type the date into the Input Field, click out or even press Enter. The value of the date should also be saved after doing so.

#### Before
before

#### After
after

### Testing
- [ ] Check that onBlurInput callback function is called and date values are stored after User typed their Date and Clicks away from component 
- [ ] Check that onBlurInput triggers onBlur Analytic 
- [ ] Check that onEnter callback function is called and date values are stored after User typed their Date and hits Enter
- [ ] Check that onEnter triggers onEnter Analytic


date-select/src/DateSelect.jsx
```diff
@@ -70,13 +70,17 @@ function validateYear({ year }) {
 }
 
 export const DATE_SELECT_ANALYTICS = { CLEAR: 'clear', NAME: 'DateSelect', SELECT: 'select' };
+export const DATE_ENTER_ANALYTICS = { CLEAR: 'clear', NAME: 'DateSelect', SELECT: 'enter' };
+export const DATE_ONBLUR_ANALYTICS = { CLEAR: 'clear', NAME: 'DateSelect', SELECT: 'onBlur' };
```
ben-burlingham: `DATE_SELECT_ANALYTICS` is one object, for all analytics on the date select.  Can you combine these, just add `ENTER` and `CLEAR` as analytics keys :)

date-select/src/DateSelect.jsx
```diff
@@ -229,35 +247,63 @@ export default class DateSelect extends React.Component {
...
+      newErrors.year = false;
     } else if (valueIsValid && !valueIsEmpty) {
-      selected = { day: value.day, month: value.month - 1, year: value.year };
-      active = { day: value.day, month: value.month - 1, year: value.year };
+      newSelected = { day: value.day, month: value.month - 1, year: value.year };
+      newActive = { day: value.day, month: value.month - 1, year: value.year };
     } else {
-      selected = { day: null, month: null, year: null };
+      newSelected = { day: null, month: null, year: null };
     }
```
ben-burlingham: This method doesn't feel right, can you explain what's going on here?

idannyou: The way we had it before mutates this.state. I just made `new${Vars}` to avoid variables all pointing to the same pointer.

Because I'm reusing these methods over and over again, I decide to just make a method to return the new variables/values in state.

date-select/sandbox/index.js
```diff
@@ -14,6 +14,8 @@ const valueFormat = ['d', 'm', 'yyyy'];
 const onAnalytics = (obj) => { console.warn('Analytics: ', obj); };
 const onClear = () => console.log('Cleared.');
 const onSelect = (v) => { console.info('Selected:', v); };
+const onBlurInput = (v) => { console.info('On Input Blur:', v); };
+const onEnter = (v) => { console.info('On Enter:', v); };
```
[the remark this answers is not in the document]
idannyou: I can make it optional? It is mostly UX, if I type in a date, some people hit enter to exit out of calendar modal.

date-select/src/DateSelect.jsx
idannyou: Done

date-select/sandbox/index.js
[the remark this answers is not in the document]
ben-burlingham: Gotcha - mostly asking since this `on*` pattern is used for passing events out of a Particle to be handled by a client.  What I was more getting at was a) is this event needed, since the particle handles closing by itself and b) should it be renamed, since "onEnter" has a different connotation than "enter was pressed".  Maybe `onExit` or `onSubmit` or something, if it's even needed at all.  Definitely should not be required though.

@nickwmurphy can you take a quick look at this to see if it aligns with the keyboard flow stuff you've been working on lately?

date-select/src/DateSelect.jsx
ben-burlingham: Ah gotcha. :+1:

ben-burlingham:
@idannyou just to confirm, are you actively using the `onSubmit` method?   If not, let's try to avoid use cases that don't exist yet 

Also, @nickwmurphy can you take a quick look here to see if this matches up with the keyboard flow work you've been doing - `tab`, `enter`, etc.  

Thanks Danny!

date-select/sandbox/index.js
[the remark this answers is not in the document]
idannyou: The Date-Select currently does close and trigger a call back ONLY after a user clicks on a date and then clicks out of the modal. 

The current Date-Select does not close and does not trigger a callback when a user types in the date, and decides to do either:
- Click out of the modal
- Or when "enter is pressed"

I think we should add those two UX functionality to this component, or at least provide it as options for users to trigger. I'll take off the isRequired for those two functionality.

idannyou:
I renamed onSubmit to onEnterPress. And made onEnterPress and onBlurInput optional.

idannyou:
@ben-burlingham I updated the README and did a version bump. Should I wait for @nickwmurphy CR before merging?

date-select/README.md
```diff
@@ -4,6 +4,11 @@ A date select.
 
 # Change Log
 
+##### 2.3.0
+- Added optional `onBlurInput` prop
+- Added optional `onEnterPress` prop
+- Added the above for an improved user experience
```
ben-burlingham: nb No need for this last bit, probalby just stick with a more technical "what was changed" not so much "why"

Issue #418, opened 2017-10-04
Version 7.5.1 breaks Table config reordering


scottwalstead:
Reordering columns in Table config no longer works in Table version 7.5.1

PR #421, opened 2017-10-04
Fix issue with updates to columnDefinitions

scottwalstead:
Fix issue with updates to columnDefinitions such as reset to default and column reordering

### Summary
There is a bug in 7.5.1 where column reordering and reset to default in Table config doesn't work because of the changes I made to columnDefinitions being in the component state and not updating with nextProps
This fixes that bug and updates the column definitions

### Related Issues
Closes: https://github.com/procore/particles/issues/418

### Testing
Try to reorder columns and click reset to default with version 7.5.2 and see if it works


scottwalstead:
Will rebase and reversion branch

table/src/Table.jsx
```diff
@@ -107,11 +107,13 @@ export default class extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
+    console.log(nextProps);
```
ben-burlingham: oops

table/README.md
```diff
@@ -4,6 +4,9 @@ A table component with self-managed checkboxes, sticky head, sticky right panel,
 
 # Change Log
 
+##### 7.5.2
```
ben-burlingham: `.2` already published

table/src/Table.jsx
scottwalstead: removed it

PR #422, opened 2017-10-04
[GLU] Tab Header with Core CSS

mboperator:
### Summary
This component introduces two major changes:
- Consuming atom level styles from `@procore/core-css` 
- A composable, declarative, stateless component pattern (similar to `Dialog`)

#### Before
screen shot 2017-10-04 at 4 29 59 pm
#### After
screen shot 2017-10-04 at 4 29 49 pm

#### Before
screen shot 2017-10-04 at 4 07 14 pm

#### After
screen shot 2017-10-04 at 4 16 20 pm 1

- The API being exposed is 95% backwards compatible to the current `Tabs` API.
  - The last 5% being that we're only merging in the _first_ stylesheet of the stylesheets array


tab-header/src/Header.jsx
```diff
@@ -0,0 +1,63 @@
...
+import Tabs, { Tab } from './Tabs';
+import HeaderLayout, { Title, Peripheral } from './HeaderLayout';
+
+import tabManagement from './tabManagement';
+import analyticsReporting from './analyticsReporting';
+
+export const HEADER_ANALYTICS = { CHANGE: 'change', NAME: 'Tabs' };
+
+// TODO: This should do something similar to compose stylesheets.
+const mergeStyles = arrayOfSheets => arrayOfSheets[0];
```
chrisarevir: You're not actually merging stylesheets here, you're actually only accessing the first one in the array?

tab-header/sandbox/index.js
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable import/no-extraneous-dependencies, no-console */
+import React from 'react';
+import { render } from 'react-dom';
+
+import ProcoreHeader, { HEADER_ANALYTICS } from '../src';
```
chrisarevir: Should maybe call it ProcoreHeading instead? Header is already the top part of the page

tab-header/src/HeaderLayout.jsx
```diff
@@ -0,0 +1,41 @@
...
+);
+
+HeaderLayout.propTypes = {
+  children: PropTypes.node,
+  style: PropTypes.shape(),
+  className: PropTypes.string
+};
+
+export const Title = ({ children, style, className }) => (
+  <div className={`${styles.slotL} ${className}`}>
```
chrisarevir: style prop isn't used here, but also this way seems inconsistent with how you used it in other files in setting style={styles} instead.

tab-header/src/HeaderLayout.jsx
```diff
@@ -0,0 +1,41 @@
...
+  </div>
+);
+
+Title.propTypes = {
+  children: PropTypes.node,
+  style: PropTypes.shape(),
+  className: PropTypes.string
+};
+
+export const Peripheral = ({ children, style, className }) => (
```
chrisarevir: style not used here either

PR #426, opened 2017-10-05
Select Classname Override Bugs

nickwmurphy:
### Summary
- Made certain that `container` and `title` have the `.expanded` class applied in the particle so that they can be overridden in the client

- New linting from `stylelint`


dannyphillips:
It was also quickly merged immediately after opening with only one +1. Is that concerning to anyone else?

PR #432, opened 2017-10-09
Invoked composeStyles in componentWillRecieveProps of inline editable components

yoyis3000:
### Summary
There is no way to update the styles of a table inline components once they are mounted. As you will see in the screenshots, this is a problem when an item becomes overdue or when an item is no longer overdue. 
In `componentWillReceiveProps` there is now a check to see if the amount of classes are the same, if not, that means the styles change and reruns `composeStyles`. 

#### Before
punchlist_before

#### After
punchlist_after

Checking to see if a class was deleted or added isn't the best solution since it doesn't take into account the possibility of changing a class. I was thinking of checking each property but that doesn't seem performant. If anyone has any suggest please let me know! 

This also assumes there is only one stylesheet passed to the component.


table/src/InlineSelect.jsx
```diff
@@ -52,6 +52,11 @@ export default class InlineSelect extends React.Component {
     if (this.props.value !== nextProps.value) {
       this.setState({ value: nextProps.value });
     }
+    const styles = Object.keys(this.props.stylesheets[0]);
+    const nextStyles = Object.keys(nextProps.stylesheets[0]);
```
dancingshell: what if you pass in more than one stylesheet?

PR #434, opened 2017-10-10
Rmf/controlled checkboxes table

RyanFuentes:
### List Partial

- Fixed bug with the slotRight prop when the filters component is not supplied.

### Paginator

- Added support for a status component that gets displayed in the middle of the pagination bar.
screen shot 2017-10-10 at 3 28 48 pm

### Table
**MAJOR BREAKING CHANGES.**

- Check boxes are no longer being controlled by the Table component's internal state.
- New props added: 'checkedRows', 'checkedGroups', and 'checkedAll' which are used to set the state of the checkboxes.
- New props added: 'onRowCheck', 'onGroupCheck', and 'onAllCheck' which are used as callbacks for when the user checks a box.
- New prop added: 'disabledGroups' which prevents users from selecting that group or the child rows.
- Bug Fix: fixed the moveHeader method from firing in case the table ever goes hidden.
- Bug Fix: fixed z-index stacking for GroupSidebarCells. This was causing issues when a dropdown list was added to a GroupSidebarCell if the next subsequent GroupSidebarCell was contained in the box of the dropdown, it would display behind the GroupSidebarCell because of the stacking order.
- Bug Fix: made a change to InlineTextEdit which will submit on blur regardless if the value had changed.


paginator/src/Paginator.jsx
```diff
@@ -144,6 +144,7 @@ export default class Paginator extends React.Component {
     return (
       <div className={this.styles.paginator}>
         <div className={this.styles.meta}>{`${msg} ${i18n.of} ${totalRecords || 0}`}</div>
+        <div className={this.styles.status}>{this.props.statusComponent}</div>
```
brock-noah: How do you feel about using children as the prop that determines what goes in the middle section?
```
<Paginator>
  <p>Some custom message</p>
</Paginator>
```

list-partial/src/ListPartial.scss
```diff
@@ -35,3 +35,8 @@
   flex-direction: row;
   justify-content: flex-end;
 }
+
+.slotRight {
+  position: absolute;
+  right: 0;
```
brock-noah: Did flex box properties not work?

paginator/src/Paginator.jsx
RyanFuentes: Using children implies the possibility that multiple children could be present and I think it's doubtful there should ever be in this situation so I'd prefer a prop.

table/src/BodyRows.jsx
```diff
@@ -4,11 +4,27 @@ import PropTypes from 'prop-types';
...
 import BodyRow from './BodyRow';
 
+
+const isGroupChecked = (group, idKey, checkedRows, checkedGroups) => {
+  if (checkedGroups) {
+    // Group checks are being controlled via props to Table
+    return checkedGroups.includes(group.key.key.toString());
+  } else {
+    return group.data.length
+      ? group.data.reduce((acc, rowData) =>
```
dancingshell: nb: `group.data.length > 0 && group.data.every(...)` might make this easer to read?

table/src/GroupRow.jsx
```diff
@@ -86,8 +87,8 @@ export default class GroupRow extends React.Component {
       }
     };
 
-    const onChange = () => {
-      onCheck(groupData.key.key);
+    const onChange = (e) => {
+      onCheck({ id: groupData.key.key, value: e.target.checked });
```
dancingshell: same comment as above with the callback signature

table/src/Table.jsx
```diff
@@ -161,57 +158,18 @@ export default class extends React.Component {
...
 
-  onRowCheck = (id) => {
-    const { checkedIds } = this.state;
-
-    checkedIds[id] = !checkedIds[id];
-
-    const allChecked = this.getAllChecked(this.props, checkedIds);
-    this.setState({ allChecked, checkedIds }, this.updateStorage);
+  onRowCheck = (e) => {
+    this.props.onRowCheck(e);
```
dancingshell: do we need to pass the whole event up?

table/src/HeadRow.jsx
```diff
@@ -125,12 +123,11 @@ export default class HeadRow extends React.Component {
     }, []);
 
     if (showCheckboxes || showGroups) {
-      const onCheckHandler = allChecked ? onUncheckAll : onCheckAll;
       const checkbox = (
         showCheckboxes
           ? (
             <label className={styles.controlCheckbox}>
-              <input type='checkbox' checked={allChecked || false} onChange={onCheckHandler} />
+              <input type='checkbox' checked={checkedAll || false} onChange={(e) => onCheckAll(e.target.checked)} />
```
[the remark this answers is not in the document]
dancingshell: FYI a new checkbox particle got merged, not sure if that is really helpful here though since it would create a cross particle dependency

table/src/BodyRows.jsx
```diff
@@ -29,7 +45,8 @@ const BodyRows = ({
     return group.data.map(rowData => (BodyRow({
       columnDefinitions,
       idKey,
-      isChecked: checkedIds[rowData[idKey]],
+      isChecked: checkedAll || checkedRows.includes(rowData[idKey].toString()),
+      isCheckDisabled: disabledGroups.includes(group.key.key),
```
dancingshell: is there any way around `key.key`? It seems like it was there before, but maybe we can make it something more descriptive?

PR #436, opened 2017-10-11
Select data prop validation

nickwmurphy:
### Summary
By allowing `keyField` and `valueField` to be passed in as props, we cannot "require" the data prop's objects to have keys named `key` and `value` without throwing lots of annoying errors.


nickwmurphy:
Accidentally branched off a branch, so commits won't make sense.

PR #439, opened 2017-10-11
Info Grid Component

dancingshell:
### Summary
Info-Grid (new component) 1.0.0
---
* New layout component `<InfoGrid/>`
* takes props `rows` (array[object]). Each item in the array produces a row, each key:value pair in the object produces a column in the row. Values can be primitives or components
* zebraStripe prop (bool) turns on and off zebra striping
```js
 rows={[
      {
        'Shark Name': 'Jaws',
        'Shark Id': '42581'
      },
      {
        'Shark Type': 'Great White',
        'Scientific Name': 'Carcharodon carcharias'
      },
      {
        Age: 4,
        Region: 'Pacific'
      },
      {
        Sightings: 33
      }
    ]}
```
screen shot 2017-10-11 at 2 22 07 pm


mcclayton:
We are still behind on making sure we have tests for all existing particles, but I think we should try to make sure that at least we make all **new** particles have tests around them.
The three npm scripts every particle should have are `sandbox`, `dev`, `prod`, and `test`: https://github.com/procore/particles#development

If you need help with this, I'd be happy to pair with you on adding some basic tests.

info-grid/src/InfoGrid.jsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+InfoGrid.propTypes = {
+  rows: PropTypes.arrayOf(PropTypes.shape()),
+  stylesheets: PropTypes.arrayOf(PropTypes.shape()),
+  zebraStripe: PropTypes.bool
+};
+
+InfoGrid.defaultProps = {
+  rows: null,
```
mcclayton: Should the default really be null instead of an empty array? There doesn't appear to be any null checks for `rows` in the code.

dancingshell: @mcclayton Good call, need to add a null check. Think'd I'd rather have the whole component return null rather than return the container with no rows in it.

info-grid/README.md
```diff
@@ -0,0 +1,8 @@
+# List Partial
```
ben-burlingham: nb oops

info-grid/src/InfoGrid.scss
```diff
@@ -0,0 +1,72 @@
...
+  display: flex;
+  flex: 1 0 200px;
+  min-width: 0;
+  border: 1px solid $gray;
+  border-width: 0 1px 1px 0;
+  padding-left: 8px;
+  align-items: center;
+}
+
+@media (max-width: 768px) {
```
ben-burlingham: nb woot

info-grid/src/InfoGrid.scss
```diff
@@ -0,0 +1,72 @@
...
+    height: 47px;
+    flex-basis: 47px;
+    width: auto;
+    &.zebra:nth-child(odd) {
+      background-color: $xLightGray;
+    }
+  }
+}
+
+//searchbar
```
ben-burlingham: nb oops

brock-noah:
Two things this component should be able to handle are
 - conditional cells
 - arrays of content (like an attachment list)

For conditional cells, I'm concerned that an array of objects isn't very open to conditionals.
We could map/filter on this array, but saving data as keys makes this difficult.

I think a more flexible approach is to provide more components that are expected children of `InfoGrid`. Then in render the developer can determine when a cell should or shouldn't be rendered.

```
<InfoGrid>
  <InfoGrid.Row>
    <InfoGrid.Label>Shark Name</>
    <InfoGrid.Value>Jaws</>
    <InfoGrid.Label>Shark Id</>
    <InfoGrid.Value>123</>
  </>
</>
```
or
```
<InfoGrid>
  <InfoGrid.Row>
    <InfoGrid.Cell>
      <InfoGrid.Label>Shark Name</>
      <InfoGrid.Value>Jaws</>
    </InfoGrid.Cell>
    <InfoGrid.Cell>
      <InfoGrid.Label>Shark Id</>
      <InfoGrid.Value>123</>
    </InfoGrid.Cell>
  </InfoGrid.Row>
</InfoGrid>
```

Then it doesn't matter the data structure of the developer is using. The dev can map over their data and render components.

This also handles arrays of data. The developer can just render each item inside the `Value` component. `InfoGrid` doesn't need to check if the typeof content is an array, it just renders what it is provided to it.

Each component is really basic, essentially just
`const Row = (props) => <div className={styles.row}>{props.children}</div>`

If you have an questions feel free to reach out to me.

ben-burlingham:

dancingshell:
@brock-noah in regards to the comment I agree that the current structure did not easily support showing and hiding cells. I added a prop `keysToHide` where you can add a remove label keys `    keysToIgnore={['Age', 'Scientific Name']}`.

For the structure part, I think for ease of getting this merged I am keeping the flatter structure for now since it fits my use case. We can always iterate towards a more nested structure as need arrises. 

Also, I wasn't really sure what you meant in regards to handling arrays of data, since currently you can pass in any component as the value, which you should be able to make work with arrays.
For example
```js
<InfoGrid
  zebraStripe
  rows={[
    {
      'Drawing No.':3
      'Is Floorplan': true
    },
    {
      Discipline: (
        <Select
          searchable
          value='Tiger Shark'
          data={[
            { key: 1, value: 'Great White Shark' },
            { key: 2, value: 'Tiger Shark' },
            { key: 3, value: 'Whale Shark' },
            { key: 4, value: 'Basking Shark' },
            { key: 5, value: 'Sharpnose Sevengill Shark and his friend Christian the Jellyfish' },
            { key: 6, value: 'Greenland Shark' },
            { key: 7, value: 'Zebra Shark' },
            { key: 8, value: 'Blacktip Reef Shark' },
            { key: 9, value: 'Cookiecutter Shark' },
            { key: 10, value: 'Frilled Shark' },
            { key: 11, value: 'Burmese Bamboo Shark' },
            { key: 12, value: 'Megamouth Shark' }
          ]} 
        />
      )
    }
  ]}
/>
```

PR #440, opened 2017-10-11
Checkbox Component

dancingshell:
### Summary
Checkbox (new component) 1.0.0
---
* straightforward checkbox component 
~~* use prop `styleVariants={['infoGrid']}` to use with info-grid~~


checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+  onCheck = () => {
+    this.setState({ isChecked: !this.state.isChecked });
+  };
+  render() {
+    const { styleVariants, stylesheets } = this.props;
+    const styles = composeStyles(baseStyles, [...getStylesheetsFromVariants(styleVariants).concat(stylesheets)]);
+    return (
+      <input
```
ben-burlingham: I'd recommend using a `<div>` here rather than a native form element, because native form elements are styled differently depending on the browser.  Those styles can't be overwritten at all.

If we are to have a true Procore checkbox, we'll need to style everything about it - borders, check style, etc.

checkbox/sandbox/sandbox.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { render } from 'react-dom';
+
+import Checkbox from '../src';
+
+render(
+  <Checkbox
+    checked
+    name='sandboxCheckbox'
```
ben-burlingham: Is there a need for a `name` prop?

chrisarevir: Form elements require a name prop for submission

dancingshell: Also for the onChange callback passes back `{value, name}`. If you have multiple checkboxes in your form this makes it easy to keep track of which one is changing

ben-burlingham: Makes sense.  We use a `scope` variable in other particles, perhaps we should change them all to `name` for consistency.

ben-burlingham: @chrisarevir yes, but that's only in a case of a `<form>` element with a child `type="submit"` button...an uncommon pattern in client side SPAs.  Also if we're making a custom checkbox so we can style it, rather than be forced to use the ugly operating system level ones, the `submit` check becomes irrelevant.

checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,63 @@
...
+    checked: PropTypes.bool.isRequired,
+    disabled: PropTypes.bool,
+    name: PropTypes.string.isRequired,
+    onChange: PropTypes.func.isRequired,
+    stylesheets: PropTypes.arrayOf(PropTypes.shape()),
+    styleVariants: PropTypes.arrayOf(PropTypes.string)
+  };
+  constructor(props) {
+    super(props);
+    this.state = { isChecked: this.props.checked };
```
mcclayton: Similar to the Text Input Component, I don't think this should be stateful.

checkbox/src/css-overrides/info-grid-checkbox-override.scss
```diff
@@ -0,0 +1,3 @@
+.container {
+  margin: 0;
```
brock-noah: I'm confused as why this override is necessary. The CSS defined on this component's base styles doesn't specify margin.

dancingshell: this gets rid of the default margin around input checkbox what came from the browser. I'll admit I only tested that in chrome

dancingshell: actually in the new way I'm implementing this, it is not necessary, removing!

dancingshell:
* prop `name` -> `scope`
* removed statefulness
* added analytics 
* updated component to not use input styling, remove need for inline variant
screen shot 2017-10-12 at 1 16 56 pm

ben-burlingham:

PR #441, opened 2017-10-11
Select 5.1.0 updates

dancingshell:
### Summary
* updated .gitignore to ignore rubymine .idea files 

Select 5.0.0 -> 5.1.0
---
* Dropdown's `renderEmpty` function is now called with current value
* Adds support for `styleVariant` prop
* Use prop `styleVariants={['infoGrid']}` to use with info-grid


.gitignore
```diff
@@ -5,3 +5,5 @@ lerna-debug.log
 node_modules
 npm-debug.log
 yarn-error.log
+.idea
+./*/.idea
```
ben-burlingham: ? no need for these?

select/src/Select.jsx
```diff
@@ -4,6 +4,7 @@ import cx from 'classnames';
 import Dropdown from './Dropdown';
 import baseStyles from './Select.scss';
 import composeStyles from '../../shared/stylesheetComposer';
+import infoGridCssOverride from './css-overrides/select-info-grid.scss';
```
ben-burlingham: Maybe this should be named something more generic - it could be used in more places than info grid and info grid might change its name in the future.  Also, `infoGrid` isn't super descriptive.  I believe this stylesheet makes it easier for inline selects, so perhaps something along those lines?

.gitignore
mcclayton: Yeah, typically editor specific files like these would go in your local .gitignore and are not committed.

select/src/Select.jsx
```diff
@@ -14,6 +15,14 @@ export const SELECT_ANALYTICS = {
   SELECT_ALL: 'select-all'
 };
 
+const variantToStylesheetMap = {
+  infoGrid: infoGridCssOverride
+};
+
+function getStylesheetsFromVariants(variantKeywords = []) {
+  return variantKeywords.map(variant => variantToStylesheetMap[variant]);
```
brock-noah: Can this be moved to the shared directory?

Possibly then create some `stylesheetComposerWithVariants` function that uses both.

dancingshell: I like this idea but I think I will wait until everything gets merged so I don't have to re-implement it 3 times

ben-burlingham:
First variant! Let's get it out and start poking at it.

PR #442, opened 2017-10-11
Text Input Component

dancingshell:
### Summary
Text-Input (new component) 1.0.0
---
* controlled text input component with clear button
* takes props `name`, `value`, `onChange`, `onSubmit`
~~* use prop `styleVariants={['infoGrid']}` to use with info-grid~~
* use prop `styleVariants={['inline']}` to use with info-grid


text-input/src/TextInput.jsx
```diff
@@ -0,0 +1,113 @@
...
+    disabled: false,
+    onChange: () => {},
+    placeholderText: '',
+    stylesheets: [],
+    styleVariants: [],
+    value: ''
+  };
+  constructor(props) {
+    super(props);
+    this.state = {
```
mcclayton: Open to discussion, but I don't believe any of these need to be in the state of this component. I think we should be favoring as stateless as possible components. These values should be passed in from a parent component functionally. If we want stateful behavior, we could add it via a HoC and still keep this base particle stateless. Thoughts?

ben-burlingham: ^ Agree, the state of text inputs is often controlled in the client anyway.

text-input/src/TextInput.jsx
```diff
@@ -0,0 +1,113 @@
...
+  };
+  render() {
+    const { disabled, stylesheets, styleVariants, placeholderText, name } = this.props;
+    const styles = composeStyles(baseStyles, [...getStylesheetsFromVariants(styleVariants).concat(stylesheets)]);
+    const clear = this.state.value
+      ? (
+        <button
+          onClick={this.onClear}
+          className={styles.clear}
+          tabIndex={-1}
```
mcclayton: Why are we taking this out of the tab order? This is generally an accessibility concern. Not sure how `a11y` focused we are, but just wanted to raise the concern. Thoughts @ben-burlingham?

ben-burlingham: Yep, this is a pattern/hack to allow `focus` and `blur` on divs.  Divs don't have those listeners by default, but by adding a `tabIndex`, they become available.

The alternative is to add (and remove) those listeners on the window object...not great.  The `tabIndex` solution manages itself, and the tab index can be whatever we wish (doesn't have to be -1).

We're not 100% on this pattern, it has its drawbacks, but the window-based blur detection isn't great either.  @nickwmurphy is working on the UX for tabbing between components so  accessibility concerns could be run by him.

ben-burlingham: ...but actually, this is a button so shouldn't need that hack? Disregard previous comment! What's this doing here? What am I doing here?

dancingshell: my understanding was that tabIndex="-1" takes something that would be tab-able and removes it from the tab order. I took this use case from the clear button on select. My guess is that when users are tabbing through inputs they may not want to stop on the clear button? Open to discussion. Also @nickwmurphy

text-input/src/TextInput.jsx
```diff
@@ -0,0 +1,116 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import baseStyles from './TextInput.scss';
+import composeStyles from '../../shared/stylesheetComposer';
+import inlineStylesOverride from './css-overrides/inline.scss';
+
+export const TEXT_INPUT_ANALYTICS = {
+  SCOPE: 'textInput',
```
ben-burlingham: NAME 😬 ...this still needs some thought lol

ben-burlingham:

text-input/src/TextInput.jsx
dancingshell: 🙃

PR #443, opened 2017-10-12
Table Inline Stylesheet Length Check

nickwmurphy:
### Summary
- Added check for stylesheet length before recomposing styles in `InlineSelect`, `InlineText`, and `InlineDateSelect`. A console error was being thrown in all cases where `componentWillReceiveProps` was being called where the `stylesheets` prop was an empty array.

- Linting


table/src/Table.jsx
```diff
@@ -320,13 +320,13 @@ export default class extends React.Component {
     });
 
     // If all columns (regardless of visibility) are autoResizable: false, warn the dev about an invalid config
-    if (columnDefinitions.every((def) => !def.autoResizable)) {
-      console.warn("Invalid column config - At least one column should not be autoResizable: false");
+    if (columnDefinitions.every(def => !def.autoResizable)) {
+      console.warn('Invalid column config - At least one column should not be autoResizable: false');
```
mcclayton: `nb` Might be clearer without the double negative:
```
Invalid column config - At least one column should be autoResizable: true
```

table/src/InlineText.jsx
```diff
@@ -37,10 +37,13 @@ export default class InlineText extends React.Component {
     if (this.props.value !== nextProps.value) {
       this.setState({ value: nextProps.value });
     }
-    const styles = Object.keys(this.props.stylesheets[0]);
-    const nextStyles = Object.keys(nextProps.stylesheets[0]);
-    if (styles.length !== nextStyles.length) {
-      this.styles = composeStyles(baseStyles, [...nextProps.stylesheets]);
+
+    if (this.props.stylesheets.length && nextProps.stylesheets.length) {
+      const styles = Object.keys(this.props.stylesheets[0]);
```
mcclayton: I don't get how this works. Is this checking for a different stylesheet by simply checking that the first stylesheet in the array has the same amount of keys as the first stylesheet in nextProps? What about two different stylesheets that have the same amount of keys but are completely different stylesheets?
What about if it were the second stylesheet in the list that changed?

yoyis3000: @mcclayton For context. Any who, I'm just gonna make a PR that invokes composeStyles in `componentWillRecieveProps` because there are so many ways a style could change.

mcclayton: Ok that'd be good since this logic as it stands is fragile.

PR #446, opened 2017-10-16
Table v8 - Export Table-Config from Table

nickwmurphy:
### Summary
1. `<TableConfig />` is now exported by `<Table />`, and is no longer a stand-alone particle. There was logic in both particles pertaining to "column definition validation." As the column definitions become more capable and complex, it will be important to have these two components share that validation logic. For now, the logic remains separated in the components, but in a future PR, I plan to extract it into a single file, unit test, or something of the sort. Eventual goal: tell devs with 100% certainty that the columns they passed in will or will not work and for what reasons.

2. `<Table />`'s `tableData` prop is now `data`, which has become the particles convention.

3. The sandbox is now similar to a real client implementation, with each column definition separated into its own file. This should improve the implementation experience by providing a sample file structure.

4. The `resizable` feature has been renamed to `userResizable` and has been removed from session storage. Storing part of a column definition in session storage proved to be a lot more than it is worth. Plus, moving forward, we would like it to be set up to persist the entire column definition in the fashion chosen by the client.

5. There is now a `default` object available in column definitions that defines the default width, hidden, and locked properties. Then when "Reset to Default" in the `<TableConfig />` is clicked, all those properties are reset to the ones listed in the default object. This means the `defaultWidth` feature was removed (and replaced by this). This also allows for columns to be hidden by default!


table/sandbox/index.js
```diff
@@ -1,144 +1,129 @@
...
 import sandboxStyles from './sandbox.scss';
-
-import columnDefinitions from './data/columnDefinitions';
-import tableData from './data/groupedData';
-// import tableData from './data/ungroupedData';
+import ungroupedData from './data/ungroupedData';
 
 console.info('SORT_DIRECTION enum:', SORT_DIRECTION);
 console.info('TABLE_ANALYTICS enum:', TABLE_ANALYTICS);
+console.info('TABLE_CONFIG_ANALYTICS enum:', TABLE_CONFIG_ANALYTICS);
```
ben-burlingham: Merge into `TABLE_ANALYTICS`?

table/sandbox/index.js
```diff
@@ -1,144 +1,129 @@
...
-        tableData
-      }}
-    />
-  </div>
-);
+        <button
+          className={sandboxStyles.sandboxButton}
+          onClick={this.handleDataChange}
+        >
+          {this.state.data === groupedData ? 'Ungrouped Data' : 'Grouped Data'}
```
ben-burlingham: :)

table/sandbox/index.js
nickwmurphy: Keeping separate for now. Not sure there is a graceful way to do that at the moment.

PR #447, opened 2017-10-16
Select variants implemented through classes/css

dancingshell:
### Summary

* fixes bug where style variants were leaking into other components due to global importing and common filename structure


dancingshell:
Closing in favor of #480

PR #448, opened 2017-10-16
Text-input variants implemented through classes/css 

dancingshell:
### Summary

* fixes bug where style variants were leaking into other components due to global importing and common filename structure


dancingshell:
closing in favor of #478

PR #452, opened 2017-10-18
Calendar

kedarmoghe:
Contributors:  Faris, Justin Watts, Kedar

This can show single day select, range select, or just the calendar itself.  It provides a onDayClick to get the clicked date.

To Do:

1.  ~~How to handle the i18n?  Right now it is hardcoded at the top, but for Procore purposes I think we have the i18n js object.  Just need to feed the correct locale to moment, ie. "en" or "es", etc.~~ .  just pass in via props

2.  ~~How to handle typed dates so it works for everyone?~~

2.  Testing (in progress)

4.  Break some of the components down (ie. Calendar.Cell) for easier to read


calendar/README.md
```diff
@@ -0,0 +1,91 @@
+# Search Bar
```
mcclayton: Copy/paste error - from Search Bar.

calendar/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@particles/date-range-select",
+  "version": "3.1.0",
```
mcclayton: Version should be `1.0.0`?

calendar/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/procore/particles/"
+  },
+  "author": "Procore",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/procore/particles/issues"
+  },
+  "homepage": "https://github.com/procore/particles/tree/master/search-bar",
```
mcclayton: ^ Link needs updating

calendar/src/Calendar.jsx
```diff
@@ -0,0 +1,485 @@
+import React from "react";
+import { withStateHandlers } from "recompose";
+import InputMask from "react-input-mask";
+import cx from "classnames";
+import moment from "moment";
+import styles from "./Calendar.scss";
+import { deepFreeze } from "lodash";
+
+/* moment setup */
+moment.locale("en"); /* get this from i18n js thingy */
```
mcclayton: Locale could simply be a prop for now, and then the client could pass that in based on the procore set locale?

calendar/src/Calendar.jsx
```diff
@@ -0,0 +1,485 @@
...
+const dateFormat = moment.localeData().longDateFormat("L");
+const weekdayLabels = moment.weekdaysMin().map(name => name.substr(0, 1));
+
+/* helpers */
+const rangeSwap = (date1, date2) => {
```
mcclayton: Since this is isolated enough logic, might help to split this out to a separate helper file?

calendar/src/Calendar.jsx
```diff
@@ -0,0 +1,485 @@
...
+  if (startDate && endDate && isValidHoveredDate) {
+    isBetweenRange =
+      date.isBetween(hoverDate, startDate) ||
+      date.isBetween(endDate, hoverDate) ||
+      date.isBetween(startDate, endDate);
+  }
+  return (
+    <div
+      onClick={() => props.onDayClick(date.utc().format())}
+      className={cx(
```
mcclayton: `nb`: Each of these does not need to be its own object. This can be compacted to:
```
cx(
  styles["calendar__day"], { 
    [styles["calendar__day--start"]]: isStartDate,
    [styles["calendar__day--not-current"]]: !isSameMonth,
    [styles["calendar__day--end"]]: isEndDate,
    [styles["calendar__day--is-between-range"]]: isBetweenRange
})
```

calendar/src/Calendar.jsx
```diff
@@ -0,0 +1,485 @@
...
+            {props.currentMonth.format("MMMM YYYY")}
+          </span>
+          <div
+            className={styles["month-navigation__next"]}
+            onClick={() => props.goToNextMonth()}
+          />
+        </div>
+        <div className={styles["calendar__container"]}>
+          <div>
+            {weekdayLabels.map((label, i) => (
```
mcclayton: Is there any more stable key we can use here than the index? Are labels not unique?

calendar/src/Calendar.jsx
```diff
@@ -0,0 +1,485 @@
...
+          <div
+            className={styles["month-navigation__previous"]}
+            onClick={() => props.goToPreviousMonth()}
+          />
+          <span className={styles["month-navigation__current-month-year"]}>
+            {props.currentMonth.format("MMMM YYYY")}
+          </span>
+          <div
+            className={styles["month-navigation__next"]}
+            onClick={() => props.goToNextMonth()}
```
mcclayton: Why not just `onClick={props.goToNextMonth}`

calendar/src/Calendar.jsx
```diff
@@ -0,0 +1,485 @@
...
+                }}
+                value={"X"}
+              />
+            </div>
+          )}
+        </div>
+        <div className={styles["month-navigation"]}>
+          <div
+            className={styles["month-navigation__previous"]}
+            onClick={() => props.goToPreviousMonth()}
```
mcclayton: I'm seeing this in a few places, this can simply be `onClick={props.goToPreviousMonth}`

calendar/src/Calendar.jsx
```diff
@@ -0,0 +1,485 @@
...
+      date.isBetween(hoverDate, endDate) || date.isBetween(endDate, hoverDate);
+  }
+  if (startDate && endDate && isValidHoveredDate) {
+    isBetweenRange =
+      date.isBetween(hoverDate, startDate) ||
+      date.isBetween(endDate, hoverDate) ||
+      date.isBetween(startDate, endDate);
+  }
+  return (
+    <div
```
mcclayton: Since this is such a large component, might be good to see if this can be broken down into smaller subcomponents. Like Calendar.Cell, etc.

calendar/src/Calendar.scss
```diff
@@ -0,0 +1,237 @@
...
+    border: 1px solid transparent;
+    border-radius: 2px;
+    box-sizing: border-box;
+    max-height: 0;
+    margin-top: -1px;
+    overflow: hidden;
+    position: absolute;
+    transition: all 0.15s ease;
+
+    // &--expanded {
```
mcclayton: Remove the comments?

calendar/src/Calendar.scss
```diff
@@ -0,0 +1,237 @@
...
+  }
+}
+
+.arrow {
+  cursor: pointer;
+  font-size: 13px;
+  text-align: center;
+
+  &::before {
+    font-family: "fontawesome";
```
mcclayton: Should we be referencing a dependency in this particle when the dependency itself is not included?

calendar/src/Calendar.jsx
kedarmoghe: since weekday label is a single character, "S" was repeated.

calendar/src/Calendar.scss
kedarmoghe: done

calendar/src/Calendar.scss
brock-noah: I agree.

calendar/src/Calendar.jsx
kedarmoghe: done

calendar/package.json
kedarmoghe: done

calendar/package.json
kedarmoghe: done

calendar/README.md
kedarmoghe: done

calendar/src/Calendar.jsx
kedarmoghe: d-   

is this a blocker if it is not in separate files?  i'm personally impartial

calendar/src/Calendar.scss
kedarmoghe: should I add a note that it needs FA ?

calendar/src/Calendar.jsx
mcclayton: Nope, not blocking 👍 Just trying to break up an already large component file.

calendar/src/Calendar.jsx
kedarmoghe: you win

calendar/src/Calendar.scss
mcclayton: I think we need something a little more contractual. Like either removing the dependency on FA or including the dependency with the particle. I'll let @ben-burlingham weigh in on this. Perhaps making icons props that can be passed in would decouple ourselves from the dependency. Just brainstorming here.

brock-noah: Haven't tried this yet, but my current idea is copy FA into particles, and have each particle compose what it needs.
```
.arrow {
  composes: fa from '../../font-awesome.css';
  composes: fa-caret from '../../font-awesome.css';
}
```
Then it isn't a requirement for the Dev to include the sheet, and each particle only loads what is necessary.

kedarmoghe: good idea!

calendar/sandbox/sandbox.jsx
```diff
@@ -0,0 +1,27 @@
...
+import sandboxStyles from "./sandbox.scss";
+
+import Calendar from "../src";
+
+const onDayClick = day => {
+  console.log(day);
+};
+
+/* isRangeToggleVisible = false,
+isRange = false */
```
ben-burlingham: Oops

calendar/sandbox/sandbox.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+render(
+  <div>
```
ben-burlingham: Why the outer div?

calendar/src/Calendar.scss
ben-burlingham: Yep, @brock-noah has the solve here

calendar/sandbox/sandbox.jsx
kedarmoghe: my bad!

calendar/sandbox/sandbox.jsx
kedarmoghe: those are just the possible props.  i will remove them

ben-burlingham:
It would be good to get this linted as well, if it's not too much trouble? `.eslintrc` in `Particles` root

kedarmoghe:
gotcha.  I've just been using prettier.  i'll set up eslint again and push the formatted code with it in a bit.

calendar/src/Calendar.jsx
kedarmoghe: I will address this in a later revision if that's fine.

calendar/sandbox/sandbox.jsx
```diff
@@ -0,0 +1,22 @@
+import { render } from "react-dom"; // eslint-disable-line
+import React from "react";
+
+import sandboxStyles from "./sandbox.scss";
+
+import Calendar from "../src";
+
+const onDayClick = day => {
```
ben-burlingham: Would recommend `onSelect` or `onChange` handlers to follow conventions in other particles, make it a touch easier to work with?

calendar/src/Calendar.jsx
mcclayton: Yep :) Perhaps consider creating a Github issue so it doesn't get lost in the mix?
This will also make unit testing this component easier when the time comes.

calendar/src/Calendar.jsx
```diff
@@ -0,0 +1,494 @@
...
+    dateRange[1] = date1;
+  } else {
+    dateRange[0] = date1;
+    dateRange[1] = date2;
+  }
+
+  return dateRange;
+};
+
+const isDateValid = (dates, index) =>
```
mcclayton: This can simply be 
```
const isDateValid = (dates, index) => !!dates[index] && Object.keys(dates[index]).length > 0
```

kedarmoghe: done

calendar/test/Calendar.test.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+const shallowRender = (props = {}) => {
+  // Merge with defaults
+  const mergedProps = Object.assign(
+    {},
+    {
+      onSelect: () => {}, // noop
+      isRangeToggleVisible: true,
+      isRange: true,
+      locale: "es"
```
mcclayton: "es" or "en"? Only asking because the default in the component seems to be "en"

kedarmoghe: wasn't sure what to put here, to be totally honest.  i will put a real test here afterwards but just put something here to make sure it passes when its given a locale.

calendar/src/Calendar.jsx
```diff
@@ -0,0 +1,494 @@
...
+  currentMonth: moment(),
+  hoverDate: {},
+  isRange,
+  isRangeToggleVisible,
+  isTyping: false,
+  keyboardTempValues: ["", ""],
+  step: 0
+});
+
+const actions = {
```
mcclayton: Not necessary I think for this PR since the scope is large, but would be nice to split this out into a presentational component and a stateful HOC around it. I think many implementers of this component would want the HOC, but splitting the two will make testing easier and will provide more power to the clients that want to manage state themselves.

kedarmoghe: i can do that later when i make the sub components

calendar/src/Calendar.scss
```diff
@@ -0,0 +1,227 @@
...
+    background: none;
+    border: 1px solid $lightGray;
+
+    &__selected {
+      background: $darkGray;
+      color: $white;
+    }
+
+    &:focus {
+      outline: 0;
```
mcclayton: `nb`: This has accessibility implications but I think this is ok for now. Seems like we don't have a large a11y push yet.

kedarmoghe: gotcha.  i will keep in mind though and fix this in the next revision!  would be cool to get this stuff on point

PR #453, opened 2017-10-18
Rmf/controlled table checkboxes

RyanFuentes:
**MAJOR BREAKING CHANGES.**

- Check boxes are no longer being controlled by the Table component's internal state.
- New props added: 'checkedRows', 'checkedGroups', and 'checkedAll' which are used to set the state of the checkboxes.
- New props added: 'onRowCheck', 'onGroupCheck', and 'onAllCheck' which are used as callbacks for when the user checks a box.
- New prop added: 'disabledGroups' which prevents users from selecting that group or the child rows.
- Bug Fix: fixed the moveHeader method from firing in case the table ever goes hidden.
- Bug Fix: fixed z-index stacking for GroupSidebarCells. This was causing issues when a dropdown list was added to a GroupSidebarCell if the next subsequent GroupSidebarCell was contained in the box of the dropdown, it would display behind the GroupSidebarCell because of the stacking order.
- Bug Fix: made a change to InlineTextEdit which will submit on blur regardless if the value had changed.


table/README.md
```diff
@@ -1,9 +1,18 @@
...
-A table component with self-managed checkboxes, sticky head, sticky right panel, and horizontal scroll widget.
+A table component with checkboxes, sticky head, sticky right panel, and horizontal scroll widget.
 
 # Change Log
 
+### 9.0.0
+- Check boxes are no longer being controlled by the Table component's internal state.
+- New props added: 'checkedRows', 'checkedGroups', and 'checkedAll' which are used to set the state of the checkboxes.
+- New props added: 'onRowCheck', 'onGroupCheck', and 'onAllCheck' which are used as callbacks for when the user checks a box.
+- New prop added: 'disabledGroups' which prevents users from selecting that group or the child rows.
```
ben-burlingham: Other particles disable entries in the data structure, a pattern I'd like to follow here: `{ key, text, children, disabled }`

This has the added benefit of being able to disable rows, and we could extend it later - what do you reckon?

table/src/HeadRow.jsx
```diff
@@ -42,10 +41,7 @@ export default class HeadRow extends React.Component {
     window.addEventListener('resize', this.debouncedResize);
 
     // Reflow violations here - make sure to only trigger once. Ben 171003
-    const controlCellWidth = (showCheckboxes || showGroups)
-      ? this.header.firstChild.offsetWidth
-      : 0;
-
+    const controlCellWidth = (showCheckboxes || showGroups) ? this.header.firstChild.offsetWidth : 0;
```
ben-burlingham: Do you have the Particles `.eslintrc` configured to the linter? IIRC this was a line length violation...no biggie though

table/src/HeadRow.jsx
```diff
@@ -95,14 +91,13 @@ export default class HeadRow extends React.Component {
 
   render() {
     const {
-      allChecked,
       allCollapsed,
+      checkedAll,
```
ben-burlingham: Recommend `isCheckedAll` to follow convention, group boolean states during declaration, and readability

table/src/GroupRow.jsx
```diff
@@ -103,8 +104,8 @@ export default class GroupRow extends React.Component {
         >
           <input
```
mcclayton: @ben-burlingham Now that we have a Checkbox component, will we want to be composing particles together? If so, how will that work when we have components implemented with different technologies?
If not, then how to we ensure consistency of styling and functionality?

table/src/Sidebar.jsx
```diff
@@ -142,15 +141,15 @@ export default class TableSidebar extends React.Component {
...
-      return <div key={`group-sidebar-${groupData.key.key}`} className={styles.groupSidebarCell}>{cell}</div>;
+      return <div key={`group-sidebar-${groupData.key.key}`} className={styles.groupSidebarCell} style={{zIndex}}>{cell}</div>;
     };
 
     const renderGroupSidebarCells = () => {
       const groups = [];
-      data.forEach((group) => {
-        groups.push(renderGroupSidebarCell(group));
+      data.forEach((group, i) => {
+        groups.push(renderGroupSidebarCell(group, groupSidebarZIndexStart + data.length - i));
```
RyanFuentes: This essentially makes the group rows descend in Z-Indexing as they go down. The reason for this is because I added a dropdrown on each Group, and because they all had the same Z-Index, the dropdown from the first group would sometimes appear behind the next group because of the stacking order.

This is why I also made the Header have a 300 Z-Index, which roughly means that any dropdowns placed in the header will appear above the Group Row unless there's > 300 groups on the page at once which seems unlikely.

table/src/Table.scss
```diff
@@ -213,11 +213,10 @@ $sidebarZ: 3;
   composes: sidebarCell;
   background: #e7e7e7;
   border-top-width: 1px;
-  z-index: 3;
+  z-index: 300;
```
RyanFuentes: See my comment in Sidebar.jsx regarding the meaning of this value.

table/README.md
RyanFuentes: This version of the table only currently supports disabling groups, which in effect, disables the groups records. I imagine we will soon add a disabledRows array which essentially does the same thing for specific rows. What changes to the drop name would you recommend?

table/src/HeadRow.jsx
RyanFuentes: Not configured but will get it setup and fix any linting issues.

table/src/HeadRow.jsx
RyanFuentes: I dig it

table/src/GroupRow.jsx
ben-burlingham: Composing particles together is something we've been able to avoid thus far.  Isolation (and simplicity) is one of the things that makes Particles easy to use and maintain.    Composition has so far been safely contained inside the Client.   

Ensuring consistency/functionality is as easy as styling the checkboxes to match the styleguide, for each Particle.  It's absolutely the job of the dev to ensure this is the case. 

We've found several times over the months that isolating Particles pays off big time, since it allows flexibility to move in the direction that the product requires without maintaining a spaghetti of dependencies (something that should happen in the Client). 

In either case, if this was composed of a Checkbox particle, changes to that Checkbox would not make it onto the client until a version bump happened. This is essentially the same as updating the table's checkbox to match the styleguide as necessary and bumping. 

A solution that would make both camps happy would perhaps be a `renderCheckbox()` prop?  But again, just an easy PR to make the checkbox follow the styleguide is probably all that's necessary here.

table/README.md
ben-burlingham: I'd suggest adding a `disabled` prop to the data structure that's passed in, so a group or a row can be `disabled`.

table/src/Sidebar.jsx
ben-burlingham: OK, these magic numbers are making me uncomfortable...could we perhaps maybe just get the length of a group array and use that?

table/src/Table.scss
ben-burlingham: In my experience large Z-indices are typically a sign that something's being shotgunned rather than surgeried...sort of a smell that "we're not sure what's going on here".  Can we make this more precise?

table/src/GroupRow.jsx
dancingshell: brainstorming here but what if the component took checkbox as a prop and then it was up to the implementing component to pass in the checkbox? Then we could modify props internally? 
```js
 const checkbox = showCheckboxes ? React.cloneElement(props.checkbox, {checked: props.isChecked}) : null
```

ben-burlingham: ^ Yep, that's what the `renderCheckbox` pattern could be :)

dancingshell: oh ya i should fully read things 👍  ✅

nickwmurphy:
Your large diff in `index.js` indicates you ran `yarn dev` before last commit instead of `yarn prod`.

table/src/Sidebar.jsx
RyanFuentes: Fixed so that it's no longer magical and will always be higher than the group headers

table/src/BodyRows.jsx
```diff
@@ -29,7 +32,8 @@ const BodyRows = ({
     return group.data.map(rowData => (BodyRow({
       columnDefinitions,
       idKey,
-      isChecked: checkedIds[rowData[idKey]],
+      isChecked: isCheckedAll || (checkedRows && !!checkedRows.find(rId => rId === rowData[idKey])),
```
evinism: nb: do we want to have row checkboxes automatically checked if isCheckedAll is true? If so, we probably want to make this consistent between both the "all" checkbox and the and "group" checkboxes, because right now group checkboxes are independent of this.

My preference would be that they're all independent, if for no other reason than it'll obscure bugs less.

table/src/GroupRow.jsx
```diff
@@ -103,8 +106,8 @@ export default class GroupRow extends React.Component {
...
             checked={isChecked || false}
-            disabled={!groupData.data.length}
-            onChange={onChange}
+            disabled={!groupData.data.length || isCheckDisabled}
+            onChange={isCheckDisabled ? () => {} : onChange}
```
evinism: tiny nitpick: we can define NOOP outside to save a function alloc.

i bet this is optimized away anyways tbh.

table/sandbox/index.js
```diff
@@ -15,6 +15,47 @@ console.info('SORT_DIRECTION enum:', SORT_DIRECTION);
 console.info('TABLE_ANALYTICS enum:', TABLE_ANALYTICS);
 console.info('TABLE_CONFIG_ANALYTICS enum:', TABLE_CONFIG_ANALYTICS);
 
+
+class TableWithStatefulCheckboxes extends React.Component {
```
evinism: were we looking to make this outwards-facing?

RyanFuentes: After this version we will be releasing the decorators. They need to be thought out a bit more and I didn't want it to delay my dependent PR.

table/src/BodyRow.jsx
```diff
@@ -47,7 +46,8 @@ const BodyRow = ({
         <input
           type='checkbox'
           checked={isChecked || false}
-          onChange={onChange}
+          disabled={isCheckDisabled}
+          onChange={isCheckDisabled ? () => {} : onChange}
```
dancingshell: nb: i think this work here because if the input is disabled `onChange` will never be called. no need for the noop function
```js
disabled={isCheckDisabled}
onChange={onChange}
```

table/src/Sidebar.jsx
```diff
@@ -12,13 +13,13 @@ const debounce = (fn, ms) => {
 };
 
 const moveHeader = (el, y) => {
-  window.requestAnimationFrame(() => {
+  el && window.requestAnimationFrame(() => {
```
RyanFuentes: this fixes a bug that occurs if the table gets hidden.

table/src/BodyRows.jsx
dancingshell: do you need to check for `checkedRows` truthiness here if you set the default prop to `[]` in the `<Table/>`?

also since you are casting the `find` call to Boolean maybe use `some` instead?

(same questions for disabledGroups and checkedGroups below)

table/src/HeadRow.jsx
```diff
@@ -128,12 +125,11 @@ export default class HeadRow extends React.Component {
     }, []);
 
     if (showCheckboxes || showGroups) {
-      const onCheckHandler = allChecked ? onUncheckAll : onCheckAll;
       const checkbox = (
         showCheckboxes
           ? (
             <label className={styles.controlCheckbox}>
-              <input type='checkbox' checked={allChecked || false} onChange={onCheckHandler} />
+              <input type='checkbox' checked={isCheckedAll || false} onChange={e => onCheckAll(e.target.checked)} />
```
dancingshell: need the false fallback here?

table/src/GroupRow.jsx
```diff
@@ -86,8 +89,8 @@ export default class GroupRow extends React.Component {
       }
     };
 
-    const onChange = () => {
-      onCheck(groupData.key.key);
+    const onChange = (e) => {
+      onCheck({ id: groupData.key.key, value: e.target.checked });
```
dancingshell: not sure if it makes sense here but i think the callbacks in other components use `{scope, value}`

table/src/GroupRow.jsx
```diff
@@ -90,7 +90,7 @@ export default class GroupRow extends React.Component {
     };
 
     const onChange = (e) => {
-      onCheck({ id: groupData.key.key, value: e.target.checked });
+      onCheck({ scope: groupData.key.key, value: e.target.checked });
```
dancingshell: would you have to change the sandbox too? 😬

table/src/GroupRow.jsx
```diff
@@ -103,8 +106,8 @@ export default class GroupRow extends React.Component {
         >
           <input
             checked={isChecked || false}
-            disabled={!groupData.data.length}
-            onChange={onChange}
+            disabled={!groupData.data.length || isCheckDisabled}
+            onChange={isCheckDisabled && onChange}
```
dancingshell: could this result in `onChange` being false instead of a function? Could probably just drop that check

PR #461, opened 2017-10-20
Info Grid style updates, plus testing! 1.1.0

dancingshell:
### Summary

* added tests
* css change where cellData should expand to fill cell. cell data content can still overwrite with it's own css width

### Testing

tests!


info-grid/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["latest", "react", "stage-0"]
```
mcclayton: Are any stage-0 features actually used by the components you're testing?

info-grid/package.json
```diff
@@ -1,12 +1,18 @@
 {
   "name": "@particles/info-grid",
-  "version": "1.0.0",
+  "version": "2.0.0",
```
mcclayton: Why is this a major bump? Did we break backwards compatibility?

info-grid/test/InfoGrid.test.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { shallow, mount } from 'enzyme';
```
mcclayton: Mount can be removed here since it isn't used.

info-grid/test/InfoGrid.test.jsx
```diff
@@ -0,0 +1,38 @@
...
+import InfoGrid from '../src';
+import InfoRow from '../src/InfoRow';
+
+const renderInfoGrid = (props = {}, renderFn = shallow) => {
```
mcclayton: `nb`: Don't need `renderFn` here since mount isn't used anywhere.

info-grid/test/InfoRow.test.jsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import { shallow, mount } from 'enzyme';
```
mcclayton: Same don't need mount here or the renderFn in `renderInfoRow` helper.

info-grid/test/InfoGrid.test.jsx
```diff
@@ -0,0 +1,38 @@
...
+    const rows = [{}, {}];
+    const infoGridWithRows = renderInfoGrid({ rows });
+
+    expect(infoGridWithRows.find('.container').children()).toHaveLength(rows.length);
+  });
+
+  it('should return only InfoRows', () => {
+    const rows = [{}, {}];
+    const infoGridWithRows = renderInfoGrid({ rows });
+    expect(infoGridWithRows.children().every(node => (
```
mcclayton: Huh... I think you are wanting `everyWhere` here: http://airbnb.io/enzyme/docs/api/ShallowWrapper/everyWhere.html

mcclayton: Then I think this can simply be: 
```
expect(infoGridWithRows.children().every(n => n.is(InfoRow));
```

info-grid/package.json
```diff
@@ -26,18 +32,24 @@
   "devDependencies": {
     "autoprefixer-loader": "^3.2.0",
     "babel-core": "^6.25.0",
+    "babel-jest": "^21.2.0",
     "babel-loader": "^7.1.1",
     "babel-plugin-transform-decorators-legacy": "^1.3.4",
     "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-env": "^1.6.0",
+    "babel-preset-es2015": "^6.24.1",
```
brock-noah: the `babel-preset-env` handles es2015+

Add `env` to the babelrc presets.

info-grid/package.json
dancingshell: I was debating this. I think the css is a "breaking" change. But I also think no one is using this yet, so I was considering still calling it `1.1.0`?

ben-burlingham: @mcclayton this is an interesting pattern: CSS changes that are breaking should bump major, but in early development, there will be a huge pile of them during iteration, so the major version will bloat.  Kaitlin and I talked about just making the breaking changes until a certain magical time when they become "important" (i.e. when a second team starts using this component?)

Just realized also that this could be a good reason for the `0.x.y` versions that @brock-noah suggested a little while ago 🤔   maybe 1.0 includes "final" CSS, among other things?

mcclayton: Sounds good to me! When I initially commented on this, I was ignorant to the fact that there were styling changes in the PR (I can't read) and was curious mainly why adding tests were a breaking change. All makes sense to me :)

dancingshell: bumping back down to `1.1.0` for now. I agree in the future we should probably go with a `0.x.y` for new components while they are in rapid development. Actually should probably change the cli tool to make that so...

dancingshell:
Updated with tests fixed. Also fixed a bug with the zebra striping in column mode, it now does not zebra stripe in column mode

info-grid/src/InfoGrid.scss
```diff
@@ -57,10 +57,10 @@
     height: 47px;
     flex-basis: 47px;
     width: auto;
+  }
 
-    &.zebra:nth-child(odd) {
-      background-color: $xLightGray;
-    }
```
dancingshell: had to remove this because now it is selecting cousin selectors so if there are not an even amount of cells in each row, this cannot be done in css 😏

PR #454, opened 2017-10-18
Testing everything

ben-burlingham:
### Summary
Putting in a testing base on:
- Bulk Editor
- Date Select
- Filters
- List Partial
- Notification Bar
- Paginator
- Tiered Select
- Select
- Spinner
- Table
- Tabs
- Uploader

### Review Notes/ Questions/ TODOs
Basic "render" on all components. Using `yarn test` interface on all.  Test output report generating as well.

### Testing
Whoa, meta.

PR #463, opened 2017-10-23
Tiered Navigation Select

brock-noah:
# Summary
A breadcrumb like picker that loads data when necessary.

Will replace the company/project picker.

## Exported Components
- `Tiers`
- `Tier`
```
<Tiers>
  <Tier>
    {(data, onDrilldown) => <span />}
  </Tier>
  <Tier>
    {(data, onDrilldown) => <span />}
  </Tier>
</Tiers>
```
API documented in [readme](https://github.com/procore/particles/blob/4419b9e806bde366fbdd1a2de539e156cd76085f/navigation-tiers/README.md)

## Internal Components
- `NavigationTiers`
The entry point. Implements the following:
- `OpenProvider`
Handles open state.
- `PageProvider`
Handles storing all pages, provides the current view's entire data list. Determines which child `Tier` is relevant. 
- `ListProvider`
Handles search term and keyboard actions. Filters the data list.
- `Dropdown`
The main view component.

- Added a shared loader for CSS
- How to add procore styles to particles? `core-react` does not import the sheet. It's done here. Should it be done here?
- Will commit built files once approved.

### Screenshots
nav tier 2


chrisarevir:
Y'all probably know this, but this functionality does exist already, we should meet up later and figure out if the use cases can be consolidated!

https://github.com/procore/particles/tree/master/tiered-select

navigation-tiers/src/Dropdown.jsx
```diff
@@ -0,0 +1,143 @@
...
+      highlightIndex,
+      isLoading,
+      placeholder,
+      searchTerm
+    } = this.props;
+
+    return (
+      <Popover placement='bottom' className={composedStyles.popover}>
+        <div className={composedStyles.searchWrap}>
+          <input
```
mboperator: We have an Input component that would work as a drop in here :)
https://procore.github.io/core/react/#input

navigation-tiers/src/NavigationTiers.jsx
```diff
@@ -0,0 +1,123 @@
...
+                      onSelect={tierProps.onSelect}
+                    >
+                      {
+                        (listData, listActions) => {
+                          if (open) {
+                            return (
+                              <div className={this.styles.positionContainer}>
+                                <Dropdown
+                                  actions={{
+                                    advance: pageActions.advance,
```
mboperator: Love this. It's clean and reads well

navigation-tiers/src/ListProvider.jsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+// const RIGHT_ARROW_KEY = 39;
```
mcclayton: Remove the comments?

navigation-tiers/sandbox/sandbox.jsx
```diff
@@ -0,0 +1,66 @@
...
+      onDrilldown={(option, cb) => {
+        // replicate request wait
+        window.setTimeout(() => cb(projectsByCompany[option.id]), 500);
+      }}
+      onSelect={onSelect}
+      placeholder='Search Companies'
+      itemSearch={(term, item) => item.name.toLowerCase().includes(term.toLowerCase())}
+    >
+      {
+        (data, onDrilldown) => (
```
mboperator: NB: I like how this reads, but it should be noted that this creates a new function on every rerender.

navigation-tiers/src/ListProvider.jsx
```diff
@@ -0,0 +1,82 @@
...
+// const LEFT_ARROW_KEY = 37;
```
mboperator: Should these be removed?

navigation-tiers/src/ListProvider.jsx
brock-noah: Yeah I'm going to ask UX about these arrow keys and would prefer not to get the key codes again.

navigation-tiers/src/PageProvider.jsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import uuid from 'uuid/v5';
+
+const BASE = '90123e1c-7512-523e-bb28-76fab9f2f73d';
```
mcclayton: Probably deserves a comment.

navigation-tiers/src/NavigationTiers.jsx
```diff
@@ -0,0 +1,123 @@
+import '@procore/core-css/build/core.css';
```
mboperator: Referencing our conversation in slack:
Maybe we should move this `import` out of the component itself and in to the `sandbox`?

Then just rely on developers on having the styles on the page. One potential downside is that we _do_ run into issues with ensuring the right version is being consumed.

navigation-tiers/src/NavigationTiers.jsx
```diff
@@ -0,0 +1,123 @@
...
+                {
+                  (tierProps, pageData, pageActions) => (
+                    <ListProvider
+                      dataIn={pageData.data}
+                      itemSearch={tierProps.itemSearch}
+                      onSelect={tierProps.onSelect}
+                    >
+                      {
+                        (listData, listActions) => {
+                          if (open) {
```
mboperator: Is there a way we can toggle a style instead of rendering `null` here? Would be a little bit cleaner and open us up to using transitions in the future.

navigation-tiers/src/PageProvider.jsx
```diff
@@ -0,0 +1,84 @@
...
+function key(arr = []) {
+  return uuid(arr.join('<(^o^)>'), BASE);
```
mboperator: Cute

navigation-tiers/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@particles/navigation-tiers",
+  "version": "0.1.0",
+  "description": "A Tiered Navigation Select Component",
+  "main": "index.js",
+  "scripts": {
+    "dev": "NODE_ENV='development' webpack --progress --colors --watch",
+    "prod": "NODE_ENV='production' webpack --progress --colors -p",
+    "sandbox": "NODE_ENV='development' webpack-dev-server --progress --colors --hot --inline ./sandbox/sandbox.jsx --config webpack.dev.config.js  --content-base ./../sandbox",
+    "test": "jest --coverage"
```
mcclayton: Can we add the most primitive test of "Component should render" to the more important of components here?
As Ben did here: https://github.com/procore/particles/pull/454

navigation-tiers/src/ListProvider.jsx
remyyounes: What about using KeyboardEvent.key instead of the deprecated KeyboardEvent.keyCode ?
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

navigation-tiers/src/NavigationTiers.jsx
mcclayton: +1

navigation-tiers/src/NavigationTiers.jsx
brock-noah: Since toggling the display property messes with the focus on mount, we decided to address this when it is a UX request.

navigation-tiers/src/Dropdown.jsx
brock-noah: Waiting for ref support.

ben-burlingham:
I'm sure you've checked it out so just FYI, there is an existing tiered select that's been run through several UX vettings, and is in use in a few places already.  

It can be used to populate a dropdown as well. 

tiered-select

mboperator:
Just noting a RL conversation here:
Because these components are `classes` for one reason or another, we're going to want to add `shouldComponentUpdate` for performance.

brock-noah:
@chrisarevir I’ll let my UX know about it

PR #464, opened 2017-10-23
Updating template

chrisarevir:
### Summary
To help with moving things over to a more reasonable build time/setup time for styling, we are moving towards utilizing postcss over node-sass for our styles. This helps setup postcss for all new particles by adding it to the CLI creator.

Also separated new components from using a shared webpack config, instead this should give all new ones their own.


_scripts/cliCreateParticle.js
```diff
@@ -48,8 +48,8 @@ function createParticle(particleName) {
 
   shell.sed('-i', 'TEMPLATE_PARTICLE', particleName, `${newParticleDir}/package.json`);
   shell.sed('-i', 'TEMPLATE_PARTICLE', particleTitle, `${newParticleDir}/README.md`);
+  shell.sed('-i', 'TEMPLATE_PARTICLE', particleTitle, `${newParticleDir}/src/index.js`);
```
dancingshell: was this a bug?

chrisarevir: Nope! I think I thought it was a bug, did a thing with it, then ended up just doing the same thing, but in a different location.

Oh, that's what I did. I alphabetized the list, but only the particleXXXX ones xD

dancingshell: I believe this it should be switched back to `newParticleComponentName` then. `particleTitle` has spaces in it, just for the read me I think at the moment.

chrisarevir:
@dancingshell done and done!

PR #468, opened 2017-10-24
Grb/filters no focus on remove

yoyis3000:
### Summary
When attempting to remove a saved filter (after a refresh) it would fetch filter options, remove the filter, add the filter back, and expand with the fetched options. The filters fetch because when you click on the remove button, onFocus fires first when it shouldn't. When its "working", the list gets expanded because of onFocus but gets removed right after because of onRemove. 

When attempting to remove a filter, it should not call on focus. 
filters


filters/src/Filters.jsx
```diff
@@ -68,6 +68,7 @@ export default class Filters extends React.Component {
 
     const [activeFilters, inactiveFilters] = getSavedActiveAndInactive(props);
     const selectedKeys = getSavedSelections(props);
+    console.log('selectedKeys', selectedKeys);
```
justinmwatts: remove

PR #470, opened 2017-10-25
Select: Unit tests.

ben-burlingham:
### Summary
Unit tests on `<Select>`.


select/src/Select.jsx
```diff
@@ -107,12 +107,12 @@ export default class Select extends React.Component {
     }
 
     /* eslint-disable no-console */
-    if (!props.i18n.clear_all) console.warn('Missing i18n.clear_all');
-    if (!props.i18n.no_items_found) console.warn('Missing i18n.no_items_found');
-    if (!props.i18n.retrieving_items) console.warn('Missing i18n.retrieving_items');
-    if (!props.i18n.search_placeholder) console.warn('Missing i18n.search_placeholder');
-    if (!props.i18n.select_all) console.warn('Missing i18n.select_all');
-    if (!props.i18n.title_placeholder) console.warn('Missing i18n.title_placeholder');
+    // if (!props.i18n.clear_all) console.warn('Missing i18n.clear_all');
```
mcclayton: Meant to delete?

select/test/Dropdown.test.jsx
```diff
@@ -0,0 +1,245 @@
...
+import baseStyles from '../src/Select.scss';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+const i18n = {
+  clear_all: 'Clear All I18N',
+  select_all: 'Select All I18N'
+};
+
+const onSelectAll = Sinon.spy();
```
mcclayton: You'll either want to create these spies in the context of the `it(` statements or reset the spy state after each test run as spies maintain state.

select/test/Dropdown.test.jsx
```diff
@@ -0,0 +1,288 @@
...
+    {
+      disabled: true,
+      key: 3,
+      value: 'ghi',
+      children: [{ key: 4, value: 'jkl' }]
+    }];
+
+    const onItemClick = Sinon.spy();
+
+    const wrapper = shallowRender({ data, onItemClick });
```
mcclayton: This is fine since the onItemClick spy is only used in one place, but I'd actually put create a new wrapper with a newly created spy all in the `it` block of the spy consumer. 
That way, if someone adds another test here that uses the shared wrapper and triggers a click function, it'd effect the run of the other test(s).

select/test/Dropdown.test.jsx
```diff
@@ -0,0 +1,288 @@
...
+  describe('ungrouped data', () => {
+    const data = [
+      { disabled: true, key: 1, value: 'abc' },
+      { key: 2, value: 'def' },
+      { key: 3, value: 'ghi' }
+    ];
+
+    const onItemClick = Sinon.spy();
+
+    const wrapper = shallowRender({ data, onItemClick });
```
mcclayton: Same as comment above

ben-burlingham: Done and done. Thanks @mcclayton

PR #471, opened 2017-10-25
Km/calendar fixes

kedarmoghe:
1.  moved some of the markup into functions that render specific pieces of the calendar

2.  added proptype validation (what do I do for moment {}?} 

3.  returns a sorted array of UTC dates now

also formatted to follow particle eslint rules.


calendar/src/Calendar.jsx
```diff
@@ -322,173 +301,288 @@ const renderDay = (props, date, isSameMonth) => {
...
+          {week.map((value, dayIndex) => (
             <div
-              className={cx(styles["range-inputs__container"], {
-                [styles["range-inputs__container--active"]]: props.step === 1
-              })}
+              key={dayIndex}
+              className={styles['calendar__day-wrap']}
+              onMouseEnter={() => setHoverDate(moment(value.date))}
+              onMouseLeave={() => setHoverDate({})}
+              onClick={() =>
```
ben-burlingham: Careful of these binds in render, perf

calendar/sandbox/sandbox.jsx
```diff
@@ -1,22 +1,20 @@
...
 
 render(
   <Calendar
     {...{
       isRangeToggleVisible: true,
       isRange: true,
-      onSelect,
-      locale: "es"
+      onSelect
+      // locale: 'fr'
```
mcclayton: Remove comment?

calendar/src/Calendar.jsx
```diff
@@ -322,173 +301,288 @@ const renderDay = (props, date, isSameMonth) => {
...
-          {props.isRange && (
+const ToggleButtons = (isRange, setIsRange) => (
+  <div className={styles['toggle-buttons']}>
+    <button
+      className={cx(styles['toggle-buttons__button'], {
+        [styles['toggle-buttons__button__selected']]: !isRange
+      })}
+      onClick={() => setIsRange(false)}
+    >
+      Single Day
```
mcclayton: `nb`: Eventually you'll probably want to i18n these

kedarmoghe: 👍

calendar/sandbox/sandbox.jsx
kedarmoghe: I put that in there so people can put whatever in here, or with this it defaults to 'en'.  I can remove it if you think it is bad to put it there.

mcclayton: It's just the sandbox so I'm fine with it 👍

calendar/src/Calendar.jsx
```diff
@@ -1,45 +1,51 @@
...
 let weekdayLabels;
 function setLocale(locale) {
   moment.locale(locale);
-  dateFormat = moment.localeData().longDateFormat("L");
+  dateFormat = moment.localeData().longDateFormat('L');
   weekdayLabels = moment.weekdaysMin().map(name => name.substr(0, 1));
 }
 
 /* helpers */
+const isValidDate = date => Object.keys(date).length > 0;
```
mboperator: NB: It'd be awesome to get unit tests started for this component. 

It might be fairly involved but there's a good amount of low hanging fruit here, IE this function :)

calendar/src/Calendar.jsx
```diff
@@ -322,173 +301,288 @@ const renderDay = (props, date, isSameMonth) => {
...
-              />
-              <button
-                className={styles["range-inputs__button-clear"]}
-                onClick={e => {
-                  props.clearDate(1);
-                  props.setStep(1);
-                }}
-                value={"X"}
-              />
+              {renderDay(
```
mboperator: Seems like this could just be another stateless component

calendar/src/Calendar.jsx
```diff
@@ -322,173 +301,288 @@ const renderDay = (props, date, isSameMonth) => {
...
+        onSelect,
+        setIsTyping,
+        setEndDate,
+        setKeyboardSingleDate,
+        setStartDate,
+        setStep,
+        step
+      )}
+      {MonthNavigation(currentMonth, goToPreviousMonth, goToNextMonth)}
+      {Calendar(
```
mboperator: Based on the way they're being used here, RangeInputs, MonthNavigation, and Calendar should be turned into stateless components

mboperator: The only change would be to wrap the arguments in an object. Swapping these functions that return JSX with stateless components is idiomatic and allows you to leverage the rest of the ecosystem 

IE: simple API and render testing becomes easy with Enzyme

calendar/src/Calendar.jsx
```diff
@@ -322,173 +301,288 @@ const renderDay = (props, date, isSameMonth) => {
...
+            setStep(1);
+          }}
+          value={'X'}
+        />
+      </div>
+    )}
+  </div>
+);
+
+const MonthNavigation = (currentMonth, goToPreviousMonth, goToNextMonth) => (
```
mboperator: NB: This should be a stateless component. Also this file is a bit long. How do you feel about splitting each component into their own file, and moving helpers to a utilities file for clarity?

calendar/src/Calendar.jsx
kedarmoghe: Will do.  That is actually going to be the next thing on this component's TODO!

calendar/src/Calendar.jsx
kedarmoghe: I can do this later if that would make the particle easier to work with.

calendar/src/Calendar.jsx
kedarmoghe: :+1: learned something new today on the why's of doing things.  thanks

PR #472, opened 2017-10-25
Km/calendar fixes

kedarmoghe:
Addressing input from @mboperator on previous PR (https://github.com/procore/particles/pull/471)

- Main reservation is the fact that you're actually using a number of stateless components but they don't fit the stateless component signature XD

- Based on the way they're being used here, RangeInputs, MonthNavigation, and Calendar should be turned into stateless components.  

- The only change would be to wrap the arguments in an object. Swapping these functions that return JSX with stateless components is idiomatic and allows you to leverage the rest of the ecosystem

Also:

- Adding tests (slowly!  I'm still learning)


calendar/src/CalendarDay.jsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import moment from 'moment';
+
+import styles from './styles.scss';
+
+const CalendarDay = ({ committedDates, date, hoverDate, isSameMonth, step }) => {
+  const startDate =
+    committedDates[0] && Object.keys(committedDates[0]).length > 0
```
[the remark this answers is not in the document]
kedarmoghe: I think because we are using moment objects, otherwise empty object.

calendar/src/CalendarParticle.jsx
```diff
@@ -0,0 +1,251 @@
...
+    const newStartDate = isValid ? moment(value, dateFormat, true) : {};
+    return {
+      ...state,
+      keyboardTempValues: [value, state.keyboardTempValues[1]],
+      committedDates: [newStartDate, state.committedDates[1]],
+      currentMonth: (isValid && moment(newStartDate, dateFormat, true)) || state.currentMonth
+    };
+  },
+  setEndDate: state => ({ value }) => {
+    const isValid = moment(value, dateFormat, true).isValid();
```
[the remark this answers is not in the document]
kedarmoghe: since this is used in the input, and date won't be valid until it matches the mask.  for example if the date to enter is 10/30/2017, and only 10/3 has been typed in, it is not valid, but the 10/3 value is held in `keyboardTempValues`

calendar/test/Calendar.test.jsx
```diff
@@ -1,30 +1,54 @@
...
+      const wrapper = renderTest(Calendar, 'mount', { isRange: true });
+      const els = wrapper.find(`.${styles['range-inputs__input']} input`);
+      expect(els.length).toBe(2);
+    });
+  });
+  describe('MonthNavigation', () => {
+    it('should render the current month', () => {
+      const wrapper = renderTest(Calendar, 'mount');
+      const el = wrapper.find(`.${styles['month-navigation__current-month-year']}`);
+      const d = new Date();
```
mcclayton: I think this is ok since it's the month, but be careful for smaller units like seconds as something like that can lead to non-deterministic test results (if the seconds become out of sync) unless we stub out date functionality.

calendar/test/Calendar.test.jsx
```diff
@@ -1,30 +1,54 @@
...
 };
 
-describe("Calendar", () => {
-  /* TODO: Placeholder test, write better ones */
-  it("should render", () => {
-    const wrapper = shallowRender();
-    expect(wrapper.exists()).toBe(true);
+describe('CalendarParticle', () => {
+  describe('ToggleButtons', () => {
+    it("should render when 'isRangeToggleVisible' is true", () => {
```
mcclayton: `nb`: I typically break out all `if` or `when` sentence fragments out into another nested describe block, just because I think it leads to clearer, more idiomatic test output.
i.e.
```
describe('CalendarParticle', () => {
  describe('ToggleButtons', () => {
    describe('when 'isRange' is true', () => {
      it("should highlight 'Date Range' button", () => {
```
Definitely a non-blocker, but just wanted to share.

mcclayton: This also allows you to group multiple `it` blocks underneath a certain pre-condition.

kedarmoghe: Thanks !  I'm still working through this part so I'll keep this in mind!

PR #475, opened 2017-10-26
Filters unit testing

ben-burlingham:
### Summary
Full battery of unit tests for Filters.


filters/src/Filters.jsx
```diff
@@ -55,7 +60,13 @@ export default class Filters extends React.Component {
 
   static defaultProps = {
     filterOptions: item => item,
-    i18n: {},
+    i18n: {
+      filter_placeholder: 'Âδδ Fïℓƭèř',
+      clear_all: 'Çℓèář Âℓℓ',
```
dannyphillips: why is this text weird? 

aka. How'd you do that?

ben-burlingham: :) 

How I did it - It's called pseudolocalization, I used this converter: http://www.pseudolocalize.com/

Why I did it - short answer, Arkadiy's i18n work. Long answer - we needed a visual cue to (strongly) suggest that an I18N candidate was missing.  `console` solutions are spammy, and `.isRequired` is too heavy handed.

filters/test/DropdownCheckboxes.test.jsx
```diff
@@ -0,0 +1,344 @@
...
+    it('is not expanded', () => {
+      expect(wrapper.state('expanded')).toBe(false);
+    });
+
+    it('is not loading', () => {
+      expect(wrapper.state('isLoading')).toBe(false);
+    });
+
+    it('triggers save handler', () => {
+      expect(onSave.called).toBe(true);
```
mcclayton: `nb` This is sort of an extraneous test since the one below also implies that the onSave was called.

filters/test/DropdownCheckboxes.test.jsx
```diff
@@ -0,0 +1,344 @@
...
+
+      const inputElement = wrapper.find(`.${styles.activeInput}`);
+      inputElement.simulate('change', { target: { value: data[0].value } });
+
+      const remainingElements = wrapper.find(`.${styles.activeItem}`);
+      expect(remainingElements.length).toBe(1);
+    });
+
+    it('displays empty message if no matching values', () => {
+
```
mcclayton: Unimplemented test

filters/test/DropdownPlain.test.jsx
```diff
@@ -0,0 +1,77 @@
...
+    styles
+  }, props);
+
+  return Enzyme.shallow(
+    <DropdownPlain {...mergedProps} />
+  );
+};
+
+describe('DropdownPlain', () => {
+  /* TODO: Placeholder test, write better ones */
```
mcclayton: Can remove comment?

filters/test/DropdownPlain.test.jsx
```diff
@@ -0,0 +1,77 @@
...
+    const titleElement = wrapper.find(`.${styles.dropdownTitle}`);
+    expect(titleElement.text()).toBe(str);
+  });
+
+  it('closes on select', () => {
+    const wrapper = shallowRender();
+    wrapper.setState({ expanded: true });
+    const itemElement = wrapper.find(`.${styles.inactiveItem}`).first();
+    itemElement.simulate('click');
+    expect(wrapper.state().expanded).toBe(false);
```
mcclayton: `nb`: This can also be written as:
```
expect(wrapper.state('expanded')).toBe(false);
```

shared/testShims.js
```diff
@@ -1,3 +1,5 @@
 global.requestAnimationFrame = (callback) => {
   setTimeout(callback, 0);
 };
+
+Object.values = obj => Object.keys(obj).map(key => obj[key]);
```
mcclayton: We should decide how we want to polyfill things like this for particles. We don't want to rely on the assumption that the client has a polyfill for this.

mcclayton: but this is exactly how we'd want to shim it for tests 👍

ben-burlingham:
Thanks @mcclayton, all resolved.  Polyfills are indeed a big assumption we should sort out at some point.

PR #476, opened 2017-10-26
Dynamic Styles

RyanFuentes:
### Summary
For context look at this : https://github.com/procore/particles/pull/438
Allows dynamic styles similar to the ticket above by adding a new prop 'DynamicStyleActive' which will allow new stylesheets to be passed in dynamically.

### Related Issues
Relates To: https://github.com/procore/particles/pull/438


ben-burlingham:
Just to note, this is still an experimental pattern - any thoughts on a better way to do this? 

I think a solution that just works without an extra prop would be cool...always compose? Compare current/next props to see if `stylesheets` is different?  Bueller?

RyanFuentes:
I think it's worthy of some stress testing to see just how much of a hit it is to adopt that pattern.

PR #477, opened 2017-10-26
added onBlur callback

RyanFuentes:
### Summary
Added actual onBlur callback. I needed this callback because the onSubmit will not fire if the user blurs and the value doesn't change.


ben-burlingham:
Funny, other consumers have intentionally added the behavior you describe.  Hopefully this solve answers everyone's needs!

PR #482, opened 2017-10-27
Select: Introducing prod shell script.

ben-burlingham:
Introducing a shell script to run the prod build process.

This will probably be extended later to include automating a changelog prompter and a semver prompter, among other things that we'd like to happen pre-publish.

It's fairly agnostic but introduces a dependency on Webpack which we can remove soon.

_Dependencies_: NPM, Webpack.

_YAT_: Shell scripting.

### Tests failed? Abort prod build.
screen shot 2017-10-27 at 12 03 55 pm

### Tests passed? Run prod build.
screen shot 2017-10-27 at 12 03 22 pm


select/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "dev": "NODE_ENV='development' webpack --progress --colors --watch",
-    "prod": "NODE_ENV='production' webpack --progress --colors -p",
+    "prod": "../_scripts/prod.sh",
     "sandbox": "NODE_ENV='development' webpack-dev-server --progress --colors --hot --inline ./sandbox/sandbox.jsx --config webpack.dev.config.js  --content-base ./../sandbox",
```
dancingshell: nb: just an idea, but you can use the pre/post npm script hooks to run the script and leave the webpack prod logic in package.json
```js
   "preprod": "../_scripts/prod.sh"
```

ben-burlingham:
@dancingshell Done.

select/package.json
ben-burlingham: Great idea - done.

PR #484, opened 2017-10-30
Select v7

nickwmurphy:
### Summary
I didn't set out to make breaking changes, but in an effort to update `titleValue` when the `titleValue` prop changed, this seemed to be a fix for many of the remaining oddities of this particle.

1. `clearable` prop added: previously, `updateOnSelect` determined whether an instance of Select was able to be cleared. This new boolean prop replaces that.

2. **MOST NOTABLE CHANGE**: `titleValue` no longer controlled by internal state: the "title value" will now be controlled by props, not a mix of props and internal state
 
3. `updateOnSelect` prop removed: now that `titleValue` is controlled outside the component, there is no need for this. 

4. `valueFunction` prop removed: now that `titleValue` is controlled outside the component, there is no need for this.

The sandbox is now a better example (IMO) of how Select is most effectively implemented in a client.


nickwmurphy:
Removed that test, as it was no longer necessary with these changes. I'll work with @ben-burlingham to see what test coverage is missing.

select/test/Select.test.jsx
```diff
@@ -68,16 +68,6 @@ describe('Select', () => {
         expect(textElement.text()).toEqual(titleValue);
       });
 
-      it('can be cleared', () => {
```
mcclayton: Isn't this test still relevant? Can we fix it instead of removing it?

select/test/Select.test.jsx
```diff
@@ -106,17 +96,28 @@ describe('Select', () => {
...
-        const wrapper = shallowRender({ titleValue });
-        const clearButtonElement = wrapper.find(`.${baseStyles.clear}`);
-        expect(clearButtonElement.exists()).toBe(true);
-      });
-
-      it('does not render a clear button if there is no title value', () => {
-        const wrapper = shallowRender();
-        const clearButtonElement = wrapper.find(`.${baseStyles.clear}`);
-        expect(clearButtonElement.exists()).toBe(false);
+      describe('has title value', () => {
```
mcclayton: `nb`
Maybe rephrase these two outer describes to be 
`when title value present`
and
`when title value not present`

select/src/Select.jsx
```diff
@@ -278,7 +268,8 @@ export default class Select extends React.Component {
 
     const title = (
       <div className={cx(this.styles.staticText,
```
mcclayton: `nb`
Minor: but this can be:
```
<div className={cx(this.styles.staticText, { 
        [this.styles.noClear]: !clearable,
        [this.styles.placeholder]: !titleValue 
})}>
```
You don't need to wrap each conditional style in an individual object. It can simply be one single object passed to classnames package.

select/src/Dropdown.jsx
```diff
@@ -136,13 +138,21 @@ export default class Dropdown extends React.Component {
 
   onClearAll = () => {
     this.props.onClearAll();
```
mcclayton: `onClearAll` is not a required prop and the default is `null` yet there is no safeguard here? This should either handle the null case or it should be a required prop.

mcclayton: Looks like this is the case for basically all of the function type props, these should be handled as well.

select/src/Select.jsx
nickwmurphy: Fixed 👍

select/src/Dropdown.jsx
ben-burlingham: @nickwmurphy lmk if you want me to help out here

ben-burlingham: That seems correct - should put in `() => {}` or similar.  But it occurs to me that this has been this way for a while and not had an issue, so React must handle it cleanly?  Anyways, `() => {}` is more correct :)

ben-burlingham: Ahh took a look at this with Nick. Seems there's several places where a prop like that is called, but there's a check to see if it's null or not.  The above case is indeed a danger point.  We'll just put in `() => {}` as defaults and remove the checks everywhere.

Nice catch!

mcclayton: 👍

PR #488, opened 2017-10-31
Update template further

chrisarevir:
Updates to the default template!


_template-particle/webpack.config.js
```diff
@@ -1,54 +1,68 @@
...
-  ]
-});
+const path = require('path');
 
+const FriendlyErrorsWebpackPlugin = require('friendly-errors-webpack-plugin');
+const WebpackNotifierPlugin = require('webpack-notifier');
 
-const path = require('path');
+module.exports = () => {
+  const prod = process.env.NODE_ENV === 'production';
```
ben-burlingham: nb Parens?

_template-particle/package.json
```diff
@@ -28,9 +28,11 @@
     "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-env": "^1.6.1",
     "babel-preset-react": "^6.24.1",
+    "babel-preset-stage-2": "^6.24.1",
```
dancingshell: I thought we were getting rid of this and moving towards using the `.babelrc` with env?

```js
{
  "plugins": ["transform-runtime", "transform-export-extensions", "transform-class-properties"],
  "presets": ["env", "react"]
}
```

_template-particle/webpack.config.js
chrisarevir: Not needed when it's a single statement and no ambiguous logic follows

_template-particle/webpack.config.js
```diff
@@ -1,54 +1,68 @@
...
+
+    entry: './src/index.js',
+
+    module: {
+      rules: [
+        {
+          test: /\.jsx?$/,
+          loader: 'babel-loader',
+          include: [path.resolve(__dirname, 'src'), path.resolve(__dirname, 'sandbox')],
+          query: { plugins: ['transform-runtime'], presets: ['env', 'react', 'stage-2'] }
```
dancingshell: remove query?

_template-particle/webpack.config.js
```diff
@@ -1,54 +1,68 @@
...
+        },
+        {
+          test: /\.s?[ca]ss$/,
+          exclude: /node_modules/,
+          use: [
+            { loader: 'style-loader' },
+            { loader: 'css-loader', options: { modules: true, importLoaders: 1, localIdentName: '[name]_[local]' } },
```
dancingshell: update to use new util https://github.com/procore/particles/blob/master/shared/cssUniqueNameBuilder.js

_template-particle/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "dev": "NODE_ENV='development' webpack --progress --colors --watch",
     "prod": "NODE_ENV='production' webpack --progress --colors -p",
-    "sandbox": "NODE_ENV='development' webpack-dev-server --progress --colors --hot --inline ./sandbox/sandbox.jsx --config webpack.dev.config.js  --content-base ./../sandbox"
+    "sandbox": "NODE_ENV='development' webpack-dev-server --progress --colors --hot --inline ./sandbox/sandbox.jsx --content-base ./../sandbox"
```
dancingshell: yay one webpack file!

_template-particle/src/TemplateParticle.jsx
```diff
@@ -3,18 +3,32 @@ import PropTypes from 'prop-types';
...
+    this.styles = composeStyles(baseStyles, props.stylesheets);
+
+    this.state = {
+      message: 'Your particle TemplateParticle says Hello World!'
+    };
+  }
+
+  render() {
+    return (
+      <div className={this.stylesheets.container}>
```
dancingshell: should be `this.styles.container`

_template-particle/package.json
chrisarevir: Easier to just commit to going stage-2 as we should want things to work completely out of the box, and most projects benefit from using stage-2

_template-particle/src/TemplateParticle.jsx
chrisarevir: Most likely, changing!

_template-particle/webpack.config.js
chrisarevir: will do once .babelrc is in use!

_template-particle/webpack.config.js
chrisarevir: On it!

_template-particle/package.json
chrisarevir: Oh, and I couldn't get it to work on my side without explicitly installing stage-2 or other ones, so I think there's some magic going on with your setup that we should make explicit for all others so it works elsewhere as well.

dancingshell: ok that works

dancingshell: thats weird, i thought we were both able to get it to work the other day with the babelrc posted above. Ya let me know if you want to pair and get to the bottom of this

PR #489, opened 2017-11-01
List partial updates

chrisarevir:
This makes the List Partial component a lot more flexible in terms of adding/removing features without needing patches (it's just a layout, should be really "dumb")

## What Changed
- Changes made to make the component more flexible to future changes.
- Breaking change: now uses `row` prop to populate different rows of the layout. An array of arrays of components.
- Breaking change: now uses `loadedArea` prop similar to `row`, but denotes components that are waiting for async data to display. E.g.: Bulk actions/paginator/table components. These are also an array of arrays of components.
- Breaking change: no longer uses `xyzComponent` props to place them in the layout


list-partial/README.md
```diff
@@ -4,6 +4,12 @@ A template partial for showing list presentations.
 
 # Change Log
 
+##### 2.0.0
+- Changes made to make the component more flexible to future changes.
+- Breaking change: now uses `row` prop to populate different rows of the layout. An array of arrays of components.
+- Breaking change: now uses `loadedArea` prop similar to `row`, but denotes components that are waiting for async data to display. E.g.: Bulk actions/paginator/table components. These are also an array of arrays of components.
+- Breaking change: no longer uses `xyzComponent` props to place them in the layout
```
ben-burlingham: @piercefrawley @evandapolarbear  Great way to call out breaking changes here! Whaddaya think? Maybe bold them or something too?

list-partial/src/ListPartial.jsx
```diff
@@ -1,103 +1,39 @@
...
-  headerComponent: PropTypes.shape(),
   isLoading: PropTypes.bool,
-  notifications: PropTypes.arrayOf(PropTypes.shape()),
-  paginationComponent: PropTypes.shape(),
-  searchComponent: PropTypes.shape(),
-  slotRight: PropTypes.shape(),
-  stylesheets: PropTypes.arrayOf(PropTypes.shape()),
-  tableComponent: PropTypes.shape()
+  loadedArea: PropTypes.arrayOf(PropTypes.shape),
+  rows: PropTypes.arrayOf(PropTypes.shape),
```
mcclayton: Is this really the most descriptive of what we expect?
Shouldn't this be more something like:
```
rows: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.node))
```
Similar comment for `loadedArea`

list-partial/src/ListPartial.jsx
```diff
@@ -1,103 +1,39 @@
...
-      </div>
+const ListPartial = ({ isLoading, loadedArea, rows, stylesheets }) => {
+  const styles = composeStyles(baseStyles, [...stylesheets]);
 
-      <div className={styles.bulkEditorComponent}>
-        { bulkEditorComponent && bulkEditorComponent }
-      </div>
+  const rowRender = rows.map(row => <div className={styles.row}>{row}</div>);
+  const loadingRender =
+    loadedArea.map(row => <div className={styles.loadedRow}>{row}</div>);
```
mcclayton: No `key`s here?

PR #491, opened 2017-11-02
[CP-593] nav tiers testing

rysmith:
### Summary
Intended to up the code coverage for the NavTier particle.  No major changes to the actual code base.

@procore/control-panel 
https://procoretech.atlassian.net/browse/CP-593

#### Before
screen shot 2017-11-02 at 10 42 50 am

#### After
screen shot 2017-11-27 at 1 22 29 pm

- The hashing function in `PageProvider` is not tested. It's current location (outside of the class) makes it hard to test.  Do we want to move it into `class PageProvider` to get better test coverage?

- `onVerticalMovement` in `ListProvider` is not tested.  I was having trouble mocking and spying on the imported function `scrollHelper`.

- `scrollHelper` is not tested.

### Testing
- [ ] Running `yarn test` in the `navigation-tiers` directory should pass all tests.


navigation-tiers/test/Dropdown.test.jsx
```diff
@@ -0,0 +1,149 @@
...
+    it('should render the search icon when there is no search term', () => {
+      const { composedStyles, wrapper } = fullRender();
+
+      expect(wrapper.find(`.${composedStyles.searchIcon}`).length).toEqual(1);
+    });
+
+    it('should render a clickable times icon when there is a search term', () => {
+      const { actions, wrapper } = fullRender({ searchTerm: 'procore' });
+
+      expect(wrapper.find('.fa-times').length).toEqual(1);
```
rysmith: need to target `composedStyles` not the font awesome class

navigation-tiers/src/OpenProvider.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import enhanceWithClickOutside from 'react-click-outside';
 
-class OpenProvider extends React.Component {
+export class OpenProvider extends React.Component {
```
rysmith: needed an export here to bypass `enhanceWithClickOutside` HoC

navigation-tiers/test/Dropdown.test.jsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+};
+
+describe('Dropdown', () => {
+
+  describe('navigation', () => {
+    it('should should render with input and navigation', () => {
+      const { wrapper } = fullRender();
+
+      expect(wrapper.find('nav').length).toEqual(1);
```
mcclayton: Typically we try to have a single assertion for every `it` statement.

navigation-tiers/test/Dropdown.test.jsx
```diff
@@ -0,0 +1,149 @@
...
+import Adapter from 'enzyme-adapter-react-16';
+import Dropdown from '../src/Dropdown';
+import Spinner from '@procore/core-react';
+
+configure({ adapter: new Adapter() });
+
+const fullRender = (props) => {
+  const requiredProps = {
+    actions: {
+      advance: spy(),
```
mcclayton: Can we make these noops `() => {}` instead of spies? 
Spies are heavier than a noop and this prevents someone from being able to reuse a wrapper between tests.
I think this would be better if you explicitly pass in the spies via mergedProps when you actually need them.

navigation-tiers/test/Dropdown.test.jsx
```diff
@@ -0,0 +1,149 @@
...
+      const { wrapper } = fullRender();
+
+      expect(wrapper.find('nav').length).toEqual(1);
+      expect(wrapper.find('input').length).toEqual(1);
+    });
+
+    it('should render the search icon when there is no search term', () => {
+      const { composedStyles, wrapper } = fullRender();
+
+      expect(wrapper.find(`.${composedStyles.searchIcon}`).length).toEqual(1);
```
mcclayton: `nb`:
This could also be written as:
`expect(wrapper.find(`.${composedStyles.searchIcon}`)).toHaveLength(1);`
or if you don't care about length and only care about existence:
`expect(wrapper.find(`.${composedStyles.searchIcon}`).exists()).toBe(true);`

navigation-tiers/test/Dropdown.test.jsx
```diff
@@ -0,0 +1,149 @@
...
+      expect(crumbs.length).toEqual(props.breadcrumbs.length);
+
+      const instance = wrapper.instance();
+
+      spy(instance.searchBarRef, 'focus');
+
+      crumbs.map((crumb, i) => {
+        crumb.simulate('click');
+
+        expect(actions.back.args[i][0]).toEqual(crumbs.length - i - 1);
```
mcclayton: This is asserting probably too much for a single test and should likely be broken down into multiple `it`s.

navigation-tiers/test/ListProvider.test.jsx
```diff
@@ -0,0 +1,161 @@
...
+import { shallow, configure } from 'enzyme';
+import { spy } from 'sinon';
+import Adapter from 'enzyme-adapter-react-16';
+import ListProvider from '../src/ListProvider';
+
+configure({ adapter: new Adapter() });
+
+const shallowRender = (props) => {
+  const requiredProps = {
+    children: spy(),
```
mcclayton: Same comment about replacing these with noops instead of spies.

navigation-tiers/test/ListProvider.test.jsx
```diff
@@ -0,0 +1,161 @@
...
+
+  describe('onChange(..)', () => {
+    it('should set the correct state values', () => {
+      const { instance } = shallowRender();
+
+      const e = {
+        target: { value: 'procore' }
+      };
+
+      expect(instance.state.searchTerm).toEqual('');
```
mcclayton: Again, I know it may seem excessive. But I think you get more idomatic, single-responsibility unit tests when you limit down your expects to a single one per `it`. In this case, maybe two are fine, but this can certainly be:
```
describe('ListProvider', () => {
...
  describe('before onChange(..)', () => {
      it('should have correct initial state...
  })

  describe('after onChange(..)', () => {
  })
```
Or maybe
```
describe('ListProvider', () => {
...
  describe('when rendered(..)', () => {
    it('should have correct initial state...
  })

  describe('after onChange(..)', () => {
     it('should set the correct state values...
  })
```

navigation-tiers/test/ListProvider.test.jsx
```diff
@@ -0,0 +1,161 @@
...
+          setHighlightIndex: instance.setHighlightIndex
+        }
+      )).toBe(true);
+    });
+  });
+
+
+  describe('onChange(..)', () => {
+    it('should set the correct state values', () => {
+      const { instance } = shallowRender();
```
mcclayton: Maybe I am missing something here, but what do we need the instance for over just using the wrapper?

navigation-tiers/test/NavigationTiers.test.jsx
```diff
@@ -1,19 +1,61 @@
...
-const shallowRender = (props = {}) => {
-  return Enzyme.shallow(
-    <NavigationTiers {...props} />
+const shallowRender = (props) => {
+  const Tier = () => <div>I'm a Tier, child of Nav</div>;
+  const requiredProps = {
+    children: [
+      <Tier
+        id='id'
+        itemSearch={spy()}
```
mcclayton: Same comment about spies vs noops.

navigation-tiers/test/OpenProvider.test.jsx
```diff
@@ -0,0 +1,60 @@
...
+    mergedProps,
+    instance
+  }
+};
+
+describe('OpenProvider', () => {
+  it('should be closed initially', () => {
+    const { instance, mergedProps } = shallowRender();
+
+    expect(instance.state.open).toBe(false);
```
mcclayton: You don't need an instance to check state, you can do this on the wrapper with:
http://airbnb.io/enzyme/docs/api/ShallowWrapper/state.html

navigation-tiers/test/ListProvider.test.jsx
```diff
@@ -0,0 +1,327 @@
...
+      highlightedItemRef: 'highlightedItemRef'
+    };
+
+    jest.mock('../src/scrollHelper.js', () => jest.fn());
+
+    const scrollHelper = require('../src/scrollHelper.js');
+
+    instance.onVerticalMovement(dropdownRef)();
+
+    it('should call scrollHelper with the correct arguments', () => {
```
mcclayton: Since this test is still in progress, perhaps denote it with `xit` instead of `it` ?

PR #492, opened 2017-11-02
Breadcrumb Particle

dancingshell:
### Summary
New Breadcrumb component!

Needed dynamic way to update breadcrumbs. I know this is much more manual way to build the breadcrumbs. Feedback welcome!

##### Non hover state
screen shot 2017-11-02 at 11 59 33 am

##### Hover state
screen shot 2017-11-02 at 11 59 28 am


breadcrumb/src/Breadcrumb.scss
```diff
@@ -0,0 +1,28 @@
...
+  }
+}
+
+.greaterThan {
+  margin: 0 6px;
+  color: rgb(103, 103, 103);
+  font-size: 13px;
+}
+
+a {
```
brock-noah: Needs to be scoped

breadcrumb/src/Breadcrumb.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+export default function Breadcrumb({ stylesheets = [], children }) {
+  const styles = composeStyles(baseStyles, stylesheets);
+  const childrenCount = React.Children.count;
+  const chevronIcon = (<i className={cx(styles.greaterThan, 'fa', 'fa-chevron-right')} />);
+  return (
+    <div className={styles.container}>
+      {React.Children.map(children, (child, index) => (
+        [
+          React.cloneElement(child, { styles, finalCrumb: index === childrenCount.length }),
```
brock-noah: `finalCrumb` doesn't seem to be used. Looks like the `active` prop on `Crumb` is used instead.

dancingshell: good call i forgot to clean this up

breadcrumb/src/Crumb.jsx
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  return (<a href={href} className={classes}>{title}</a>);
+}
+
+Crumb.propTypes = {
+  title: PropTypes.string.isRequired,
+  href: PropTypes.string,
+  active: PropTypes.bool,
+  styles: PropTypes.object
```
mcclayton: Should this be required?
Won't this error if no styles are passed in?

breadcrumb/test/Breadcrumb.test.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { configure, shallow, mount } from 'enzyme';
```
mcclayton: `nb`: `mount` is not used.

breadcrumb/src/Breadcrumb.scss
```diff
@@ -0,0 +1,28 @@
+@import '../../shared/variables.scss';
```
mcclayton: Doesn't look like any variables are actually referenced here.

breadcrumb/src/Crumb.jsx
dancingshell: It would but since I am attaching the styles prop to the Crumb in Breadcrumb through `React.cloneElement`, it should always have them. Suggestions welcome on a better way to do this, but since on the implementer side styles aren't required, I didn't want to throw an error

breadcrumb/src/Breadcrumb.jsx
```diff
@@ -0,0 +1,29 @@
...
+export default function Breadcrumb({ stylesheets = [], children }) {
+  const styles = composeStyles(baseStyles, stylesheets);
+  const childrenCount = React.Children.count;
+  const chevronIcon = (<i className={cx(styles.greaterThan, 'fa', 'fa-chevron-right')} />);
+  return (
+    <div className={styles.container}>
+      {React.Children.map(children, (child, index) => (
+        [
+          React.cloneElement(child, { styles }),
+          index < childrenCount.length && chevronIcon
```
brock-noah: `nb` I'm not sure how this is working?
`React.Children.count(children)` should return a number, but you don't pass it children and call length instead? `React.Children.count.length`

dancingshell: Ah so right! I think it was somehow only working with 3 children 🤦‍♀️ . Updated!

PR #493, opened 2017-11-02
[Navigation Tiers] Arrow Keys Scrolling

brock-noah:
### Summary
- When navigating with the arrow keys, if your item is out of the container, it will scroll with you.

### Screenshots
scroll

Refs and animatedScrollTo mimic existing company/project header.


navigation-tiers/src/scrollHelper.js
```diff
@@ -0,0 +1,52 @@
...
+  t /= d/2;
+  if (t < 1) return c/2*t*t + b;
+  t--;
+  return -c/2 * (t*(t-2) - 1) + b;
+};
+
+const animatedScrollTo = (element, to, duration, callback) => {
+  const start = element.scrollTop,
+  change = to - start,
+  animationStart = +new Date();
```
mcclayton: `nb`: I know this was taken from another package and I am not sure how often this function is called, but 
`new Date.getTime()` is more performant than `+new Date()`.
https://stackoverflow.com/a/14255732

mcclayton: Oh, if this is on arrow key press then this is not important enough for perf. Disregard.

PR #494, opened 2017-11-02
[Select 9.0.0] Flat, N-Level Grouped, and Hybrid Lists

piercefrawley:
### Summary
- Support for n level child grouping
- Adds back internal search held in component state for both flat and nested list
 (https://github.com/procore/particles/pull/508)
- Render logic will now be controlled by prop`displayGroupedChildren`
- Controlled/Uncontrolled search is contolled by prop `shouldFilter`

#### Flat
screen shot 2017-11-14 at 6 02 22 pm

#### Grouped
screen shot 2017-11-14 at 6 02 38 pm

#### Hybrid
screen shot 2017-11-14 at 6 02 50 pm


mcclayton:
Looks like some tests are failing with this change, could you please fix those up?

mcclayton:
I don't think we want to commit both the package.lock here. There is already a yarn.lock file.

select/src/Dropdown.jsx
```diff
@@ -43,6 +44,7 @@ export default class Dropdown extends React.Component {
   static defaultProps = {
     closeOnSelect: true,
     data: [],
+    displayGroupChildren: true,
```
evandapolarbear: Do we want to default this to true?

select/src/Select.scss
```diff
@@ -285,13 +285,32 @@ $h: 30px;
...
+  padding-left: 42px;
+}
+
+.pad2 {
+  padding-left: 64px;
+}
+
+.pad3 {
+  padding-left: 86px;
 }
```
ben-burlingham: nb Good candidate for `@for`?

select/sandbox/data.js
```diff
@@ -1,69 +1,38 @@
...
+      { key: 2, value: 'B' },
+      { key: 3, value: 'C' }
     ]
   },
-  { value: 'Kids Games',
-    key: 2,
-    disabled: true,
+  {
+    key: 8,
+    value: 'Group 2',
```
ben-burlingham: 😢 all that nice test data

select/src/Dropdown.jsx
piercefrawley: nope! planning on keeping it false

select/sandbox/data.js
piercefrawley: totally agree! was just using faux data for the grouping. I'll bring the board games back and come up with some appropriate subcategories

select/src/Select.scss
piercefrawley: exactly what I was looking for, thanks!

select/src/Dropdown.jsx
```diff
@@ -215,32 +218,45 @@ export default class Dropdown extends React.Component {
...
         <div
-          className={cx(styles.item, styles.groupItem,
+          className={cx(styles.item, styles.groupItem, styles[`pad${item.depth}`],
             { [styles.disabled]: item.disabled,
               [styles.focusedItem]: focusedItem === i,
+              [styles.childItem]: !item.children,
               [styles.lastItem]: ungroupedData.length && array.length - 1 === i })}
           data-id={`group-${item[keyField]}`}
-          key={`group-${item[keyField]}`}
+          key={`group-${item[keyField]}-${Math.random(10)}`}
```
dancingshell: `Math.random()` shouldn't be used for key identifiers as this will cause unintended mounting and unmounting of the component. Keys should be the same each time you iterate in order for react efficiently re-render.
https://reactjs.org/docs/reconciliation.html#tradeoffs

select/src/Dropdown.jsx
```diff
@@ -215,32 +218,45 @@ export default class Dropdown extends React.Component {
...
-    const groupedData = filteredData.reduce((acc, item) => {
+    const group = collection => collection.reduce((acc, item) => {
       if (item.children) acc.push(item);
       return acc;
     }, []);
 
+    const groupNested = (collection, acc, n) => collection.reduce((a, item) => {
+      let depth = n;
+      acc.push(Object.assign(item, { depth }, {}));
+      if (item.children) {
```
dancingshell: `item.children.length`? not sure if you have tried it with empty child array

select/src/Dropdown.jsx
```diff
@@ -215,32 +218,45 @@ export default class Dropdown extends React.Component {
...
+        groupNested(item.children, acc, n + 1);
+        depth = 0;
+      }
+      return acc;
+    }, []);
```
dancingshell: might be reading this wrong, but do you need to pass in `acc`? or if you want to pick the starting value using `acc` set `},  acc)` on line 233 which will init `a` which is what should be passing around the reduce fn instead of `acc`?

select/src/Dropdown.jsx
```diff
@@ -215,32 +218,45 @@ export default class Dropdown extends React.Component {
...
-    const groupedItems = groupedData.length
-      ? groupedData.map((item, i, array) => (
+    const groupedItems = groupedData.length ?
+      groupedData.map((item, i, array) => (
         <div
-          className={cx(styles.item, styles.groupItem,
+          className={cx(styles.item, styles.groupItem, styles[`pad${item.depth}`],
             { [styles.disabled]: item.disabled,
               [styles.focusedItem]: focusedItem === i,
+              [styles.childItem]: !item.children,
```
dancingshell: same thing with length here

PR #495, opened 2017-11-02
[Navigation Tiers] Fix scroll issue when hovering

brock-noah:
### Summary
Cannot listen to prop changes because item hover also changes the highlighted index.

### Related Issues
https://twitter.com/bradurani/status/862462407800741888

#### Before
scroll bug


navigation-tiers/src/ListProvider.jsx
```diff
@@ -50,12 +66,25 @@ export default class ListProvider extends React.Component {
     }
   }
 
+  onVerticalMovement = (list, item) => {
```
mcclayton: Do we need to do any checks here for null refs? This will error if you the search text entered returns no results.

PR #496, opened 2017-11-02
Select custom search/filter 7.1.0

dancingshell:
### Summary
* Allow for custom filter to be passed in to filter dropdown results. Default filter can be combined with custom filter to create result set
* Css fix for inline disabled hover state

### Note
This was the best way I could think of to pass around all the parts of the data needed to filter + use the default filter, open to other interface suggestions


ben-burlingham:
Hi @dancingshell, I think the search functionality already exists using `onSearch` to filter the `data` on the client.  Does this do something new?

dancingshell:
Talked with @ben-burlingham in person, moving towards making the filtered data passed in as data prop and removing the filtering from the dropdown/select components

select/src/Dropdown.jsx
```diff
@@ -4,6 +4,10 @@ import cx from 'classnames';
 
 import { SELECT_ANALYTICS } from './Select';
 
+export function defaultSearchFn({ value = '', valueField = Dropdown.defaultProps.valueField, data }) {
+  return data.filter(item => item[valueField].toLowerCase().includes(value.toLowerCase() || ''));
```
mcclayton: `includes()` does not have very good browser support and requires polyfills for a number of browser versions.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

PR #497, opened 2017-11-02
Table stickyHeader option

tonynd:
### Summary
- Allow users to disable sticky headers on the table.

- Defaults to sticky header.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "particles",
-  "version": "2.0.0",
+  "version": "2.1.0",
```
ben-burlingham: Looks like you ran `npm version minor` in the `particles` root - should be in `particles/table` :)

ben-burlingham:
Addressed in https://github.com/procore/particles/pull/527

PR #500, opened 2017-11-03
SearchBar: Style updates.

ben-burlingham:
### Summary
- UI/UX overhaul
- Preprod unit test runner added
- Changed `.searchContainer` class to `.container`
- Removed Font Awesome in favor of inline SVG

screen shot 2017-11-03 at 6 42 54 am


search-bar/src/SearchBar.jsx
```diff
@@ -75,27 +105,35 @@ export default class SearchBar extends React.Component {
   }
 
   render() {
+    const clear = this.state.value && (<div
+      className={cx(this.styles.btnClear)}
```
mcclayton: `nb`: Don't really need to use `cx` here for a single class

ben-burlingham: :+1: fixed

PR #502, opened 2017-11-03
Paperwork

chrisarevir:
Small css changes to match UX/Product wants


bulk-editor/README.md
```diff
@@ -1,5 +1,8 @@
 # Bulk Editor
 
+##### 2.0.2
+- Changed default styling to align more with product/UX desires
```
ben-burlingham: nb Suggest more impersonal verbage "UI/UX update" or "Default styling change" or something

PR #503, opened 2017-11-03
Small changes to list-partial sandbox

chrisarevir:
General layout-ing, Official layout design as of 11/2/17


chrisarevir:
It's actually not! Since  it's just sandbox changes, there's no prod diff!

mcclayton:
D'oh! I need 👓

PR #508, opened 2017-11-03
Select 8.0.0, uncontrolled filtering

dancingshell:
### Summary
Previous implementation https://github.com/procore/particles/pull/496 closed in favor of this PR

##### 8.0.0
 - _Breaking_: Dropdown no longer filters, implementing component must now pass in filtered data via `data` prop
 - _Breaking_: `onSearch` is now called with `{value, scope}`
 - Previous filter function is now exported for convenience as `defaultSearchFn`
 - Fixed inline variant disabled hover style


select/src/Dropdown.jsx
```diff
@@ -4,6 +4,10 @@ import cx from 'classnames';
 
 import { SELECT_ANALYTICS } from './Select';
 
+export function defaultSearchFn({ value = '', valueField = Dropdown.defaultProps.valueField, data }) {
+  return data.filter(item => item[valueField].toLowerCase().includes(value.toLowerCase() || ''));
```
mcclayton: `includes()` does not have very good browser support and requires polyfills for a number of browser versions.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

dancingshell: We were using this before in the code, I just moved it up the file... are you suggesting we shim or find an alternative?

mcclayton: Yeah, I know that you did not introduce the function, but I am saying that we should either change to implementation or include a polyfill within the particle (not one that replaces the String prototype method., but perhaps that can simply be imported from a _shared util file).
Wrench has a polyfill included for `.includes` but I don't think we should be making assumptions that the clients will include the appropriate polyfills.
cc @ben-burlingham

thewillhuang: webpack should handle it, but its not setup right, look into https://github.com/babel/babel/tree/master/experimental/babel-preset-env#forcealltransforms to do complete transformations so you are safe on all browsers

piercefrawley: Could you swap for `indexOf`?

ben-burlingham: No thoughts on polyfills yet, seems like @thewillhuang is onto something there or yeah as @piercefrawley just build it so it doesn't need a polyfill...would be nice to keep things as minimal as possible.

dancingshell: hey all just checked and seems we are currently using `.includes` in 2 other places in addition to this:
- https://github.com/procore/particles/blob/master/table/src/InlineSelect.jsx#L189
- https://github.com/procore/particles/blob/master/info-grid/src/InfoRow.jsx#L9

So going forward we could change all of them to use `indexOf` or implement some webpack config, which might be more flexible going forward

select/sandbox/index.js
```diff
@@ -34,6 +36,13 @@ export default class Sandbox extends React.Component {
   renderGroup = group => group.value;
   renderItem = item => item.value;
 
+  onSearch = ({ value, scope }) => this.setState({ searchTerm: value });
+
+  filterDisciplines = () => {
```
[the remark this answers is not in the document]
dancingshell: fixed!

select/src/Dropdown.jsx
dancingshell: Updated this PR to use `indexOf` and made an issue for solving the `Array.prototype.includes` issue https://github.com/procore/particles/issues/519

dancingshell:
new additions to this PR
 - _Breaking_: Clicking the bottom slot now closes the dropdown
 - _Breaking_: Closing the dropdown now clears the previous search term
 - Selected item now has highlighted style for ungrouped data
 - Disabled dropdown if data is loading
 - (more) Fixed inline variant disabled hover style

PR #509, opened 2017-11-03
[Navigation Tiers] Start on N

brock-noah:
### Summary
Allows the component to be mounted and not start at the root page.
Why? On project level the user needs to start on projects not company.

### Screenshots
tier n

`onDrilldown` and `onGet` are very similar. They both get data for a tier. There could be a single `onFetch`, and the render prop of `onDrilldown` would take the next nodes `onFetch`.

The main difference is `onDrilldown` prop provides the entire object while `onGet` only returns the id. These could be quickly consolidated if only returning the id is acceptable, while a little more work is required for the object case.


navigation-tiers/README.md
```diff
@@ -74,6 +91,12 @@ Provides a function that will have an option of the current tier. Determine how
 
 ## Change Log
 
+### 0.4.0
+- BREAKING CHANGE - rename `Tiers.props.data` to `Tiers.props.initialData` to show intention of mounting with this data
```
rysmith: `nb` Should we bump the major because of the breaking change or do we not want to move to 1.0 yet?

mcclayton: I think the idea is that, for a new particle, as we get up and running we are technically in an experiment/iterate phase where we will introduce several back to back breaking changes and will lead to version bloat early on. 
This convention keeps us at a `0.x.x` version until we deem the component to be "stable" and bump it to its 1.0.0 version.

brock-noah: What MC said
http://semver.org/#spec-item-4
http://semver.org/#doesnt-this-discourage-rapid-development-and-fast-iteration

PR #511, opened 2017-11-06
Sticky-Footer v2

nickwmurphy:
### Summary
- Match UI mocks
- Removed `content` prop to instead support `children`
- Fixed positioning bug
- Update to react 16 and react-dom 16
- Classname is now `.stickyFooter`, not `.footer`

### Screenshot
screen shot 2017-11-05 at 8 12 27 pm


nickwmurphy:
I think this particle would be a little heavy handed for just a footer. I would be inclined to wait for the use case on that.

PR #512, opened 2017-11-06
Filters Key Listeners

nickwmurphy:
### Summary
- Up arrow, down arrow, enter, esc, tab key listeners added
- Clear all removed
- Select all is now a checkbox
- Linting

### Screenshot
screen shot 2017-11-05 at 8 52 03 pm

**Note:** I tried writing the `onKeyDown`, `scrollDown`, and `scrollUp` methods in a more generic fashion and passing them down from `Filters.jsx` as props, but it was not very elegant. We could consider a refactor to consolidate that code into a single place, but for now, this feels like the best solution.


filters/src/FilterAvailable.jsx
```diff
@@ -64,25 +59,93 @@ export default class FilterAvailable extends React.Component {
 
   onChange = (evt) => {
     this.setState({ expanded: false });
-    this.props.onChange(evt);
+    let key;
+
+    if (evt.key === 'Enter') {
```
ben-burlingham: nb might recommend `keycode` (13) rather than magic string, seems like it would be more future-proof, don't recall offhand. Readability is nice with these though 🤔

filters/test/Filters.test.html
```diff
@@ -14,15 +14,13 @@
...
 	.suite-table td.test .failureMsg {font-weight:bold;font-size:0.7rem;}
 	.suite-table td.result {width:20%;text-align:right;}
-</style></head><body><h1>Filters</h1></body><div id="timestamp">Start: 2017-11-3 14:26:01</div><div id="summary">
-		46 tests /
-		46 passed /
-		0 failed /
+</style></head><body><h1>Filters</h1></body><div id="timestamp">Start: 2017-11-4 19:00:30</div><div id="summary">
+		15 tests /
+		13 passed /
+		2 failed /
```
chrisarevir: 2 Failing tests :(

filters/src/FilterCheckbox.scss
```diff
@@ -65,7 +49,7 @@
 }
 
 .filterCheckboxSearch {
-  padding: 16px;
+  padding: 16px 16px 0;
```
ben-burlingham: nb I personally prefer the longhand over the "three" shorthand...maybe just something I have to get used to

filters/test/Filters.test.html
ben-burlingham: Yikes!

filters/src/FilterAvailable.jsx
brock-noah: `keycode` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode). I believe `key` or `code` are the replacements.

filters/src/FilterCheckbox.jsx
```diff
@@ -169,32 +179,105 @@ export default class FilterCheckbox extends React.Component {
...
+    this.setState({ focusedItem });
+  };
+
+  onHoverItem = item => this.setState({ focusedItem: item });
+
+  scrollDown = () => {
+    if (this.tokenBodyScroll) {
+      const items = [...this.tokenBodyScroll.childNodes].slice(0, this.state.focusedItem + 2);
+      const scrollPosition = items.reduce((acc, item) =>
+        acc + item.clientHeight, 0) - this.tokenBodyScroll.clientHeight;
```
brock-noah: It would be nice not to reduce for scrolling. Can we use this value to determine the top of the highlighted item? `this.state.focusedItem.offsetTop`

filters/src/FilterCheckbox.jsx
```diff
@@ -169,32 +179,105 @@ export default class FilterCheckbox extends React.Component {
...
+      if (scrollPosition >= this.tokenBodyScroll.scrollTop) {
+        this.tokenBodyScroll.scrollTop = scrollPosition;
+      }
+    }
+  }
+
+  scrollUp = () => {
+    if (this.tokenBodyScroll) {
+      const items = [...this.tokenBodyScroll.childNodes].slice(0, this.state.focusedItem - 1);
+      const scrollPosition = items.reduce((acc, item) => acc + item.clientHeight, 0);
```
brock-noah: Same comment as above.

filters/src/FilterCheckbox.jsx
```diff
@@ -169,32 +179,105 @@ export default class FilterCheckbox extends React.Component {
...
 
-  onClearSearch = (evt) => {
-    evt.stopPropagation();
-    this.searchInput.value = '';
-    this.searchInput.focus();
-    this.setState({ hiddenKeys: {}, value: '' });
+  onClearSearch = () => {
+    if (this.searchInput) {
+      this.searchInput.focus();
+      this.searchInput.value = '';
```
brock-noah: Why are we explicitly setting the search value?

brock-noah: nvm we aren't controlling the search input's state.

filters/src/FilterAvailable.jsx
ben-burlingham: :themoreyouknow:

filters/src/FilterCheckbox.jsx
nickwmurphy: I'll look at optimizing that in a follow up. I like the idea.

brock-noah: You also have the index of the focused child, and childNodes as an array.

PR #514, opened 2017-11-06
Recompose styles on `componentWillReceiveProps`

nickwmurphy:
### Summary
- Recompose styles on `componentWillReceiveProps`
- Talked to @yoyis3000 about this (he has encountered this in a different particle)...seems the complexity of deciding if the stylesheets prop _did_ change is not worth the effort. After attempting to do so myself, I agree.


nickwmurphy:
Performance concerns are valid...maybe we can revisit this again soon.

brock-noah:
Why is the stylesheet changing? This seems like an anti pattern.

PR #516, opened 2017-11-06
Template Particle: Adding test skeleton.

ben-burlingham:
### Summary
Adding testing pattern to TemplateParticle.

(This does not create a required dependency for new Particles, only ones that opt in to this common template.)


_template-particle/package.json
```diff
@@ -6,7 +6,22 @@
   "scripts": {
     "dev": "NODE_ENV='development' webpack --progress --colors --watch",
     "prod": "NODE_ENV='production' webpack --progress --colors -p",
-    "sandbox": "NODE_ENV='development' webpack-dev-server --progress --colors --hot --inline ./sandbox/sandbox.jsx --content-base ./../sandbox"
+    "preprod": "../_scripts/preprod.sh",
+    "sandbox": "NODE_ENV='development' webpack-dev-server --progress --colors --hot --inline ./sandbox/sandbox.jsx --content-base ./../sandbox",
+    "test": "jest --coverage"
```
dancingshell: add verbose option for reporting with multiple test files?

_template-particle/test/TemplateParticle.test.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import Enzyme from 'enzyme';
+import Sinon from 'sinon';
+import Adapter from 'enzyme-adapter-react-16';
+import TemplateParticle from '../src';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+const shallowRender = (props = {}) => {
+  const mergedProps = Object.assign({}, {}, props);
```
dancingshell: double empty object?

_template-particle/README.md
```diff
@@ -4,5 +4,5 @@ Describe your Particle here!
 
 # Change Log
 
-### 1.0.0
+### 0.1.0
```
dancingshell: thanks

_template-particle/src/index.js
```diff
@@ -1 +1,3 @@
-export default from './TEMPLATE_PARTICLE';
+import TemplateParticle from './TemplateParticle';
```
dancingshell: this will break the sed I believe

_template-particle/test/TemplateParticle.test.jsx
ben-burlingham: Left the second one there since it has to assign to an empty object...maybe I'll put in a bogus prop to illustrate what should be happening.

_scripts/cliCreateParticle.js
```diff
@@ -47,11 +47,16 @@ function createParticle(particleName) {
   shell.cp('-R', './_template-particle/*', newParticleDir);
 
   shell.sed('-i', 'TEMPLATE_PARTICLE', newParticleComponentName, `${newParticleDir}/sandbox/sandbox.jsx`);
-  shell.sed('-i', 'TEMPLATE_PARTICLE', newParticleComponentName, `${newParticleDir}/src/index.js`);
-  shell.sed('-i', 'TEMPLATE_PARTICLE', particleName, `${newParticleDir}/package.json`);
+  shell.sed('-i', /TEMPLATE_PARTICLE/g, newParticleComponentName, `${newParticleDir}/src/index.js`);
+  shell.sed('-i', 'TEMPLATE_PARTICLE', newParticleComponentName, `${newParticleDir}/test/TemplateParticle.test.jsx`);
```
dancingshell: doe we need the global replace here too?

ben-burlingham: Sure, it's global per line.  I've gone ahead and just put all the patterns to match multiple occurrences per line.

_template-particle/test/TemplateParticle.test.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import Enzyme from 'enzyme';
+import Sinon from 'sinon';
```
mcclayton: Sinon is not used here anywhere, and I'm not sure we should include it in the dependencies by default.
Jest has spies, albeit much more limited than sinon. So I think we should leave that dependency up to particle implementers to pull in on their own.

PR #520, opened 2017-11-07
Feedback banner v 0.1.1

Jacky-Lei:
### Summary
- Title, tooltip text, button text, and button link can be configured through props. Thanks @mboperator for the `Button` and `Tooltip` atoms!

### Related Issues
- Built to be used initially for https://procoretech.atlassian.net/browse/CP-535

### Screenshots
screen shot 2017-11-06 at 11 26 30 pm

- Check in with UX about positioning of tooltip
- Check in with UX about button (size sm) no matching mock (https://procore.invisionapp.com/d/main/#/console/12148499/262028563/inspect)

### Testing
- [ ] Tooltip works
- [ ] Button links correctly to forum


ben-burlingham:
I'm sure y'all have your own reasons for spending time to build another notification bar, just dropping by to mention that a `notification-bar` already exists, with a `body` and a `title` prop that can compose Atoms in a Procore Client, rather than introducing a dependency on Atoms in Particles.

ben-burlingham:
I'm sure y'all have your own reasons for spending time to build another notification bar, just dropping by to mention that a notification-bar already exists, with a body and a title prop that can compose Atoms in a Procore Client, rather than introducing a dependency on react-core in Particles.

Jacky-Lei:
@ben-burlingham I can see `notification-bar` making sense to make a feedback banner given you can pass in the info-icon with the tooltip and button through the body and pass in `noDismiss` to get rid of the times icon. The styling is a tad different around the borders, padding, and margins but that could've been adapted per this banner's needs too. 

I'm down to redo this using the `notification-bar` but wonder if we should have a `FeedbackBanner` as its own particle so when future squads don't have to replicate code as this seems like a UX standard.

ben-burlingham:
So the good news here is that we're moving to ship Procore styles, in Particles.  The border/padding/etc stuff that you mention will be a wonderful contribution to the existing notification bar, so everyone could use those styles. 

`<FeedbackBanner>` could indeed merit its own Particle as well - since it's not quite a notification bar, you're probably right that it needs its own. 

Either way, your call of course, just wanted to let you know of the existing options.  Thanks for considering it and LMK if you want help with unit tests or existing Particle patterns.

feedback-banner/src/FeedbackBanner.scss
```diff
@@ -0,0 +1,36 @@
...
+  padding: 0 24px;
+  font-size: 15px;
+  font-weight: bold;
+}
+
+.info-icon {
+  position: relative;
+  margin-left: 8px;
+
+  .tooltip-container {
```
ben-burlingham: nb Since styles are scoped, it's usually a good idea to keep the specificity as low as possible, so people can override easily.  Nesting things makes them just a touch harder to override, so I usually steer clear of that structure in this environment.

feedback-banner/README.md
```diff
@@ -0,0 +1,9 @@
+# Feedback Banner
+
+Feedback banner builder that comes with tooltip and button.
+Title, tooltip text, button text, and button link can be configured through props.
+
+# Change Log
+
+### 1.0.0
```
dancingshell: this should be `0.1.0` to match package.json. this was a bug in the create script that since has been fixed

feedback-banner/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@particles/feedback-banner",
+  "version": "0.1.0",
+  "description": "Description of your new Particle",
```
dancingshell: add short description please. Hopefully in the near future we will integrate a description into the new particle cli

feedback-banner/src/FeedbackBanner.scss
```diff
@@ -0,0 +1,36 @@
...
+  color: $white;
+  background: $info-blue;
+  font-family: Roboto, 'Open Sans', Lato, sans-serif, Arial;
+  height: 32px;
+  padding: 0 24px;
+  font-size: 15px;
+  font-weight: bold;
+}
+
+.info-icon {
```
dancingshell: nb: camelCase css names

ben-burlingham: There's an SCSS linter available in the root

feedback-banner/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "url": "https://github.com:procore/particles/"
+  },
+  "author": "Procore",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/procore/particles/issues"
+  },
+  "homepage": "https://github.com/procore/particles/tree/master/feedback-banner",
+  "dependencies": {
+    "@procore/core-css": "^2.3.14",
```
brock-noah: dev dependency?

feedback-banner/sandbox/sandbox.jsx
```diff
@@ -0,0 +1,15 @@
...
+import React from 'react';
+import { render } from 'react-dom';
+import FeedbackBanner from '../src';
+
+render(
+  <FeedbackBanner
+    title='Beta'
+    tooltipText='Some details about Beta'
+    buttonText='Submit Your Feedback'
+    buttonLink='https://procore.uservoice.com/forums/905677-feedback-for-my-procore'
```
brock-noah: Should probably use a fake link in the sandbox.

feedback-banner/src/FeedbackBanner.jsx
```diff
@@ -0,0 +1,57 @@
...
+            title={tooltipText}
+          />
+        </div>
+      </span>
+      <Button
+        size='sm'
+        variant='info-outline'
+        href={buttonLink}
+        className={styles.button}
+        target='blank'
```
brock-noah: `_blank`

feedback-banner/src/FeedbackBanner.scss
```diff
@@ -0,0 +1,40 @@
...
+.infoIcon {
+  position: relative;
+  margin-left: 8px;
+
+  &:hover {
+    cursor: pointer;
+
+    .tooltipContainer {
+      display: block;
+      font-family: Roboto, 'Open Sans', Lato, sans-serif, Arial;
```
brock-noah: is this necessary? does it not inherit from `banner`

Jacky-Lei: Unfortunately, it inherits from the fa-icon, which has `font-family: FontAwesome`

feedback-banner/package.json
Jacky-Lei: @farism @mboperator Is it ok to have this in devDependency and force users of this package to have `@procore/core-css` installed on their own project?

ben-burlingham: Particles are fully isolated, so it's best as not a dev dependency since it's a requirement for this Particle.

feedback-banner/src/FeedbackBanner.scss
brock-noah: 👍  Didn't notice it was inside the span

feedback-banner/package.json
brock-noah: My concern is this line will add global css, say `core-spinner` for this particle and in our app we import it, again adding a global of `core-spinner`. If every particle imported core styles, it could be a tossup for what `style` tag is last in the DOM. Though if there is one import, then particles could be using a different version. Not sure what to do here. Keep it in until it's a problem?

PR #522, opened 2017-11-07
Spinner: Unit testing.

ben-burlingham:
### Summary
- _Breaking_ Using pseudolocalization for i18n handling
- _Breaking_ Removing `className`
- Many more unit tests added
- Added preprod hook

screen shot 2017-11-07 at 8 50 55 am


spinner/test/Spinner.test.jsx
```diff
@@ -1,24 +1,49 @@
...
 describe('Spinner', () => {
-  /* TODO: Placeholder test, write better ones */
-  it('should render', () => {
+  it('wraps child elements', () => {
+    const children = <div>foo-child</div>;
+    const wrapper = shallowRender({ children });
+    expect(wrapper.contains(children)).toBe(true);
+  });
+
+  it('does not have loading class if isLoading is false', () => {
```
mcclayton: `nb`:
I like to break out "if" and "when" conditions that are in `it` statements into a nested describe for clearer, more idiomatic test output.

i.e.
```
describe('Spinner', () => {
    describe('when `isLoading` is false', () => {
        it('does not have loading class', () => {
```
Definitely not blocking though, nor do you need to make the change. Just my personal preference as I think it leads to clearer more self-documenting tests.

ben-burlingham: Yeah, that felt funny - thanks for the tip. Will do.

spinner/test/Spinner.test.jsx
```diff
@@ -22,16 +22,20 @@ describe('Spinner', () => {
...
+      const asyncLoadElement = wrapper.find(`.${styles.asyncLoad}`);
+      expect(asyncLoadElement.hasClass(styles.loading)).toBe(false);
+    });
   });
 
-  it('has loading class if isLoading is true', () => {
-    const wrapper = shallowRender({ isLoading: false });
-    const asyncLoadElement = wrapper.find(`.${styles.asyncLoad}`);
-    expect(asyncLoadElement.hasClass(styles.loading)).toBe(false);
+  describe('if isLoading is false', () => {
```
mcclayton: I think you meant for this to be `if isLoading is true`. Also, this is a duplicate test of the one above - I think you forgot to flip the boolean values.

PR #523, opened 2017-11-07
Readme updates

ben-burlingham:
### Summary
README was getting a bit cluttered.  I've cleaned it up significantly so newcomers can get started more easily, and removed references to patterns that are not Particles-wide.


README.md
```diff
@@ -3,80 +3,48 @@
...
-> Particle names must be unique and contain only letters and '-' dashes
+### How can I create a new Particle?
 1. In the root dir run `yarn create-particle <your-particle-name>`
 
 # Development
 
 To develop a component `FOO`, all you need are these commands:
 
-### `yarn sandbox`
+#### `npm run sandbox`
```
mcclayton: If we are changing this to `npm`, do you want to change the command to create a new particle as well?
Also, there are a number of other places in the README that reference `yarn`.
Also, any reason we are changing this from `yarn` to `npm`?

README.md
```diff
@@ -3,80 +3,48 @@
...
 
-### `yarn test`
+#### `npm run test`
 
 Starts the appropriate test runner for `FOO`'s test suite and runs the corresponding test file(s) from within the `/test` directory.
 Also generates a code coverage report for `FOO` component and prints it to the console.
 
-# Testing
-
-All Storybook documented `particles` get a thin layer of snapshot testing provided by `Storyshots` and jest.
```
mcclayton: This seems fine to me. We could move this to specific particles' READMEs that follow this convention if this is something we still want.

README.md
ben-burlingham: Yep, sounds good.

Re: changing from `yarn` to `npm` - officially, we've been able to have only NPM as a dependency.  Having `yarn` here suggests that's a dep as well, which is not the case.

README.md
```diff
@@ -3,80 +3,48 @@
...
 1. In the root dir run `yarn create-particle <your-particle-name>`
 
 # Development
 
 To develop a component `FOO`, all you need are these commands:
 
-### `yarn sandbox`
+#### `npm run sandbox`
 
 Spins up `FOO` and serves it up on `http://localhost:8080` (or `http://ude.8080`) for you to play around with in a standalone environment. The sandbox files are usually minimal implementation examples for the component, so keep them up to date!
```
mcclayton: Do we still want to include `http://ude.8080` ?
I assume this refers to our UDE setup, however - for an open source repo, and for no other information on the UDE here - should we just remove it?

README.md
ben-burlingham: My concern here is that not all Particles support storybook, and that it was a temporary solution while documentation was getting off the ground.  Now that we have prod hooks and testing patterns, storyshots becomes less relevant.  Also, since we have a common, agnostic documentation pattern now, storybook is less relevant. 

So a per-particle solution might be best here.  I'd love to keep the Readme simple so the barrier to entry is not intimidating, so leaning toward leaving it out for now unless there's strong feelings that it should stay.

README.md
mcclayton: Only concern will be ensuring that whatever each particle chooses to use, that they only commit either the `package.lock` or `yarn.lock` file. Perhaps we can have Danger Bot or something check to ensure we don't commit both for a given particle.
Thoughts?

ben-burlingham: Interesting - that sounds like a good solve. No thoughts about that yet so up for anything that works!

mcclayton: Creating a Github Issue for trackability. 👍

chrisarevir: Agree with this, but harder to implement unless there's a standard across particles

README.md
```diff
@@ -3,80 +3,48 @@
...
-- MINOR version when you add functionality in a backwards-compatible manner
-- PATCH version when you make backwards-compatible bug fixes.
+# Publishing Checklist
+- Make sure the sandbox works.
+- Update the README with each change
+- `npm version patch|minor|major` in the component directory:
+  - `major` version when you make incompatible API changes,
+  - `minor` version when you add functionality in a backwards-compatible manner
+  - `patch` version when you make backwards-compatible bug fixes.
+- Run `yarn prod` to build a minified prod bundle
```
chrisarevir: still references yarn

chrisarevir:
If we switch the readme to use yarn, then we should probably also enforce that we are all in on npm, or all in on yarn as they each have a lock file now, and they are not compatible. NPM is stable and has long history, Yarn now has history and is fast, but we have to enforce one or the other. That being said, I think the change to the README is perfectly fine until we find a way/agree on which one. That or we dangerbot out the whole thing and it's one or the other for everything. :shrug:

ben-burlingham:
Sounds good.  I also just realized we have a pile of dependencies in the particle root, so it's not entirely true to say that NPM is our only dep.  We should work on pruning those down in conjunction with dangerbot setup.

PR #525, opened 2017-11-07
Tabs unit tests

ben-burlingham:
### Summary
- _Breaking_ Changing `data` prop from `{ k, v }` to `{ key, value }`
- _Breaking_ Requiring `onChange`
- Making `selectedKey` proptype either number or string
- Adding unit tests, CSS module mocking, and code coverage reporting.
- Adding preprod hook

screen shot 2017-11-07 at 3 03 16 pm


tabs/package.json
```diff
@@ -57,6 +58,7 @@
     "qs": "^6.3.0",
```
mcclayton: `qs` Is no longer used by this particle and can be removed.

tabs/sandbox/sandbox.jsx
```diff
@@ -1,5 +1,39 @@
...
+const onChange = (key, value, index) => { console.warn(`onChange: { key: ${key}, value: ${value} }, index: ${index}`); };
+
+const data = [
+  { key: 'foo', value: 'The Foo' },
+  { key: 'bar', value: 'A Bar' },
+  { key: 'baz', value: 'Some Baz' }
+];
+
+const scope = 'sandbox-tabs';
+const selectedKey = data[0].k;
```
mcclayton: Should this be `data[0].key` ?

tabs/test/Tabs.test.jsx
```diff
@@ -17,9 +22,69 @@ const shallowRender = (props = {}) => {
 };
 
 describe('Tabs', () => {
-  /* TODO: Placeholder test, write better ones */
-  it('should render', () => {
-    const wrapper = shallowRender();
-    expect(wrapper.exists()).toBe(true);
+  it('renders n tabs', () => {
```
mcclayton: Maybe change this to say `it('renders tab for every data entry', () => {`
then the test can be:
```
it('renders tab for every data entry', () => {
    const wrapper = shallowRender({ data });
    const tabElements = wrapper.find(`.${styles.tab}`);
    expect(tabElements.length).toBe(data.length);
});
```
I think this is a bit more contractual and more descriptive of what we actually expect. However, it then makes it slightly more brittle if someone modifies data. So I'd almost then decide to inline data in the `it` statement if reuse of that fixture becomes too heavy. Just some thoughts.

tabs/test/Tabs.test.jsx
```diff
@@ -17,9 +22,69 @@ const shallowRender = (props = {}) => {
...
+  });
+
+  describe('active tab', () => {
+    it('uses selected key prop', () => {
+      const wrapper = shallowRender({ data, selectedKey: data[1].key });
+      const activeElement = wrapper.find(`.${styles.active}`);
+      expect(activeElement.text()).toBe(data[1].value);
+    });
+
+    it('does not trigger callback', () => {
```
mcclayton: Perhaps `does not trigger onChange` ?
Perhaps even more descriptive like:
```
describe('active tab', () => {
    describe('when clicked', () => {
        it('does not trigger onChange', () => {
        });
    });
});
```

tabs/test/Tabs.test.jsx
```diff
@@ -17,9 +22,69 @@ const shallowRender = (props = {}) => {
...
+      const onChange = Sinon.spy();
+      const wrapper = shallowRender({ data, selectedKey: data[1].key, onChange });
+      const activeElement = wrapper.find(`.${styles.active}`);
+      activeElement.simulate('click');
+      expect(onChange.called).toBe(false);
+    });
+
+    it('does not trigger analytics', () => {
```
mcclayton: Similar structural comment as above.

tabs/src/Tabs.jsx
```diff
@@ -14,17 +15,16 @@ class Tabs extends React.Component {
       }),
```
mcclayton: Need to update the shape of data in the proptypes here:
Should be:
```
data: PropTypes.arrayOf(
    PropTypes.shape({
        key: PropTypes.string,
        value: PropTypes.string
    })
)
```

tabs/test/Tabs.test.jsx
ben-burlingham: Much better!

tabs/test/Tabs.test.jsx
ben-burlingham: I was thinking that myself, but backed away since `onChange` is not necessarily a permanent name.  I'd like the test to indicate that a callback is called, but not necessarily introduce a naming dependency.

mcclayton: I just like to be more explicit, as `callback is called` does not provide much meaningful description apart from viewing the test itself.
I want the test output to be self-documenting of the behavior itself so that at a glance - a developer can tell what the component does without having to look at either the component or the test code.

mcclayton: It's fine if `onChange` is not a final name. If the name is changed, the test will have to be updated anyway - so might as well call out the name in the test description as well as it will get changed if updated as well.

ben-burlingham: Makes sense - fixed.  I'll follow in other tests as well then.

tabs/test/Tabs.test.jsx
```diff
@@ -17,9 +22,73 @@ const shallowRender = (props = {}) => {
...
+  it('renders tab for every data entry', () => {
+    const wrapper = shallowRender({ data });
+    const tabElements = wrapper.find(`.${styles.tab}`);
+    expect(tabElements.length).toBe(data.length);
+  });
+
+  describe('active tab', () => {
+    describe('when clicked', () => {
+      it('uses selected key prop', () => {
+        const wrapper = shallowRender({ data, selectedKey: data[1].key });
```
mcclayton: This one does not rely on there being a click event, so maybe this one should be moved outside the `when clicked` describe.

ben-burlingham: OK it's time for a break, I'm going too fast! Thanks (again) and fixed.

PR #531, opened 2017-11-07
Added tableRow onClick functionality to the tables

tonynd:
### Summary
- Allow users to add a functionality when clicking an entire table row.

- Useful when all the columns represent 1 entity.
- Insights would like the entire row to be clickable, and redirected to the corresponding reports page.


table/README.md
```diff
@@ -4,6 +4,9 @@ A table component with checkboxes, sticky head, sticky right panel, and horizont
 
 # Change Log
 
+##### 9.2.3
+- `handleTableRowOnClick` prop added. Allows an onClick functionality when clicking the whole row.
```
ben-burlingham: Please change to `onRowClick` to follow the naming convention.

table/sandbox/index.js
```diff
@@ -146,6 +146,7 @@ export default class Sandbox extends React.Component {
           columnDefinitions={this.state.columns}
           data={this.state.data}
           globalKey='table-sandbox'
+          handleTableRowOnClick={() => {}}
```
ben-burlingham: The sandbox is meant to illustrate what happens when a row is clicked.  Can this be more informative?  An example might look like:

`onRowClick={(rowId) => { console.warn('Row clicked: ', rowId); }}`

You can check out other sandboxes for more examples.

table/README.md
ben-burlingham: Also, this is new functionality, so should be a `minor` bump (not a `patch`)

ben-burlingham: Also, looks like this version doesn't agree with the `package.json` one (should be a `minor` since new functionality)

table/sandbox/index.js
tonynd: @ben-burlingham thanks Ben. Just pushed the new changes.

PR #533, opened 2017-11-08
Table: Inline select and date select use portals!

dancingshell:
### Summary

Dropdowns for inline components were pretty broken in small table datasets or near the bottom of the table. This PR takes advantage of the new react portal feature in react 16
- Updated to React 16.x.x
- `InlineSelect` and `InlineDateSelect` now use react portals for better popout positioning
- rearranged functions a bunch in `InlineDateSelect` to make it easier to read
- tests now pass

### Related Issues
Closes:   
Relates To: 

#### Before
screen shot 2017-11-07 at 4 42 23 pm

#### After
screen shot 2017-11-07 at 4 42 00 pm

Since the portal content (dropdown content) is now no longer nested in the parent container class, some styles need to be re-implemented in the dropdown itself


table/README.md
```diff
@@ -4,11 +4,16 @@ A table component with checkboxes, sticky head, sticky right panel, and horizont
...
+##### 9.4.1
+- Updated to React 16.x.x
+- `InlineSelect` and `InlineDateSelect` now use react portals for better popout positioning
+
 ##### 9.4.0
 - `onRowClick` prop added. Allows an onClick functionality when clicking the whole row.
 
 ##### 9.2.2
 - `stickyHeader` prop added
+>>>>>>> master
```
mcclayton: Looks like an artifact of a merge resolution?

table/src/InlineDateSelect.scss
```diff
@@ -114,6 +114,11 @@
     border-radius: 2px;
     right: 0;
   }
+
+  * {
+    box-sizing: border-box;
+    font: 11px sans-serif;
```
ben-burlingham: nb Might not want to put a font declaration in the universal selector; it's quite hard to override and will probably cause a mismatched font error in any production stack

table/src/InlinePortal.jsx
```diff
@@ -0,0 +1,45 @@
...
+      this.createPortal(nextProps);
+    }
+    if (nextProps.show === false && this.props.show === true) {
+      this.closePortal();
+    }
+  }
+  componentWillUnmount() {
+    this.closePortal();
+  }
+  createPortal(props) {
```
ben-burlingham: nb linting?

table/src/Table.scss
```diff
@@ -1,6 +1,6 @@
 @import '../../shared/variables';
 
-$headerZ: 2;
+$headerZ: 6;
```
ben-burlingham: nb Is this a carefully calculated value, and not a shotgun?

chrisarevir:
The only issue I see with this is that in order to take advantage of any of the new portal stuff is that the client needs to be on react16 for it to work. Particles do not ship with their own version of react so it might end up breaking/not working properly if a client is not "up to date". See the shared webpack.config.js at the root and notice the "external" key names react and react-dom.

table/package.json
```diff
@@ -1,13 +1,13 @@
 {
   "name": "@particles/table",
-  "version": "9.4.0",
+  "version": "9.4.1",
```
chrisarevir: This would be a breaking change if now on react 16 since the API for the needed parts is no longer there.

table/src/InlineDateSelect.jsx
```diff
@@ -137,16 +139,15 @@ export default class InlineDateSelect extends React.Component {
     }
     this.styles = composeStyles(baseStyles, [...nextProps.stylesheets]);
   }
-
   onBlur = () => {
     document.activeElement.blur();
     this.setState({ expanded: false });
-  }
+  };
```
chrisarevir: Needed?

table/src/InlineDateSelect.jsx
```diff
@@ -296,7 +285,7 @@ export default class InlineDateSelect extends React.Component {
 
     this.setLeft(dropdownLeft, tableLeft);
     this.setTop(dropdownTop, tableBottom, tableTop);
-  }
+  };
```
chrisarevir: Needed?

dancingshell:
Updated to major version bump. Also included:
- Date picker now opens to initial value if passed in via prop and highlights date
- Fixed bug where clicking inline date title didn't open picker

table/src/InlineDateSelect.scss
dancingshell: even when it is nested inside another class?

ben-burlingham: That arguably makes it worse, unfortunately, because anyone who needs to override it will have to try harder.  Star selectors are generally evil for a variety of reasons - performance, specificity, and lack of precision.

Issue #519, opened 2017-11-07
Ensure Array.prototype.includes works on all browsers


dancingshell:
Based on discussion from https://github.com/procore/particles/pull/508#discussion_r148843134, we need to make sure our configs are set up to support this method if we are using it.

ben-burlingham:
This is a larger conversation about polyfills - sort of a no-brainer.  We need to add polyfills.  

IE11 and up?

ryanyogan:
https://github.com/zloirock/core-js  Perhaps a light-weight drop-in solution, for now, would do the trick?

PR #537, opened 2017-11-08
Added percent width support to table column definitions

tonynd:
### Summary
- Allow users to specify `px` or `%` of width in the column definitions.

- Does not break current configuration. Defaults to `px` if none is given.


table/src/BodyRow.jsx
```diff
@@ -21,14 +31,15 @@ const BodyRow = ({
...
+    const maxWidth = !def.autoResizable ? { maxWidth: `${definitionWidth}` } : {};
 
     return (<div
       key={`cell-${rowData[idKey]}-${def.key}`}
       className={cx(styles.tableCell)}
       style={{
-        flex: `1 0 ${def.width}px`,
-        minWidth: `${def.width}px`,
+        flex: `1 0 ${definitionWidth}`,
+        minWidth: `${definitionWidth}`,
```
dancingshell: need template string?

table/src/BodyRow.jsx
```diff
@@ -2,6 +2,16 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 
+const getWidth = (def) => {
+  switch (typeof def.width) {
+    case 'string':
+      return (def.width.includes('px') || def.width.includes('%')) ?
```
dancingshell: We were just discussing in another PR that `.includes` is not well supported on browsers and we don't properly polyfill it at the moment. Maybe switch on `indexOf` instead. See #519 

Also, nb: Switch might be a bit overkill here. Since there is just one case with a ternary in it, might be simplified with an if statement. Another option is this function can either return `%` or `px` from here to concat onto width below

PR #538, opened 2017-11-09
Table - Fix passing the rowData to onRowClick

tonynd:
### Summary
- Fix passing rowData to the `onClick` functionality.

- Allow access to the rowData when clicking on a row.


brock-noah:
@rysmith this should be helpful

PR #539, opened 2017-11-09
[Navigation Tiers] Large List Performance

brock-noah:
### Summary
Performance lags when lists are large because it has to map the dataset during render. Solution is to limit the map size.
Tried out [react-virtual-list](https://github.com/developerdizzle/react-virtual-list), issues with holding arrow keys the page would jump and pass large chucks and issue when wrapping from the top -> bottom or visa-verse because those elements aren't there.

### Change Log
- fix clear button position from global button tag css targeting
- active breadcrumb style
- limit search results list length
  - prop for upper limit
  - prop for message when too large

### Screenshots
screen shot 2017-11-08 at 4 47 57 pm


navigation-tiers/src/ListProvider.jsx
```diff
@@ -5,6 +5,7 @@ import scrollHelper from './scrollHelper';
 export default class ListProvider extends React.Component {
   static propTypes = {
     children: PropTypes.func.isRequired,
+    clipResultsAt: PropTypes.number.isRequired,
```
Jacky-Lei: Should this be required?

brock-noah: It is in this component because the parent sets it. Everyone is going to have this problem so I think it should be required. It can be set to something high like 10000 if desired.

PR #546, opened 2017-11-11
Filters calendar i18n fixes

justinmwatts:
### Summary
- Fixes i18n bugs for the calendar filter. The calendar filter input range now supports i18n date formats.


filters/src/Filters.jsx
```diff
@@ -98,11 +98,14 @@ export default class Filters extends React.Component {
...
-      no_items: 'Nô ïƭè₥ƨ áƲáïℓáβℓè ƒôř ƭλïƨ ƒïℓƭèř.',
-      select_all: '§èℓèçƭ Âℓℓ'
+      filter_placeholder: 'Add Filter',
+      clear_all: 'Clear All',
+      clear_all_filters: 'Clear All',
+      no_items: 'No items available for this filter.',
+      select_all: 'Select All',
+      single_day: 'Single Day',
+      date_range: 'Date Range',
+      locale: 'en'
```
ben-burlingham: The pseudolocalization is an intentional pattern to gracefully prompt the dev that I18N is a requirement, without spamming the console.

http://www.pseudolocalize.com/ or any other tool can provide the strings to restore these.

justinmwatts: ahh, good to know. I'll get this updated.

filters/src/Filters.jsx
```diff
@@ -102,7 +102,10 @@ export default class Filters extends React.Component {
       clear_all: 'Çℓèář Âℓℓ',
       clear_all_filters: 'Çℓèář Âℓℓ',
       no_items: 'Nô ïƭè₥ƨ áƲáïℓáβℓè ƒôř ƭλïƨ ƒïℓƭèř.',
-      select_all: '§èℓèçƭ Âℓℓ'
+      select_all: '§èℓèçƭ Âℓℓ',
+      single_day: '§ïñϱℓè Ðá¥',
+      date_range: 'Ðáƭè Ráñϱè',
+      locale: 'en'
```
ben-burlingham: nb Does `en` need to be in I18N (is it displayed somewhere)?  Seems like it should be a separate `locale` prop?

justinmwatts: it's not displaying anything. I can make it a separate prop.

ben-burlingham: There is an argument for keeping it in i18n, but typically that structure is used for strings that are to be displayed.  Doesn't matter much

PR #547, opened 2017-11-11
CoachMarks v-0.1.0

Jacky-Lei:
https://procoretech.atlassian.net/browse/CP-619

### Summary
- Overlay takes CoachMarks as children and accepts callback as prop to handle click
- CoachMarks can be fixed or attached to element through props

### Screenshots
screen shot 2017-11-10 at 4 28 59 pm

- react-css-modules for core-css?
- move some CSS variables like `$bg-color: #259DF8`, `$sm: 13px;`, `$bolder: 900;` to shared?
- Next version: move css to 1 stylesheet and remove `CoachMark` as open component but allow user to pass in data structure as prop to `Overlay` and built out CoachMark internally


ben-burlingham:
Are you providing a way to override styles?

Jacky-Lei:
@ben-burlingham thanks :) Each component takes a stylesheet like other particles to override the styling

coach-marks/src/Overlay.scss
```diff
@@ -0,0 +1,34 @@
+@import '../../shared/variables.scss';
+
+$bold: 700;
+$md: 14px;
```
brock-noah: Are these variables or the imported variables used?

coach-marks/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## usage
+
+- Overlay takes CoachMarks as children and accepts callback as prop to handle click
+- CoachMarks can be fixed or attached to element through props
+
+# Change Log
+
+### 0.1.0
+- Initial commit
```
brock-noah: It would be nice if this described the initial features. Like accepts numbers for static pinning or an element id for dynamic pinning.

coach-marks/src/Overlay.scss
```diff
@@ -0,0 +1,34 @@
...
+
+.veil {
+  position: fixed;
+  top: 0;
+  left: 0;
+  background: rgba(0, 0, 0, 0.3);
+  width: 100%;
+  height: 100%;
```
brock-noah: def `nb` fyi
`position: fixed||absolute; top: 0; left: 0; width: 100%; height: 100%` is the same as
`position: fixed||absolute; top: 0; left: 0; bottom: 0; right:0`

coach-marks/src/Overlay.scss
Jacky-Lei: These are the ones that were imported previously from my-procore's shared/variables.

ben-burlingham:
@Jacky-Lei awesome!  Just FYI the other way that I've found is useful when composing children together is to just compose the styles once at the root component, then pass a required `styles` prop to any children, who then consume that.  Advantages here are only compose once, and less imports to maintain :)

coach-marks/src/CoachMark.jsx
```diff
@@ -0,0 +1,116 @@
...
+      right: `${props.fixedRight}px`,
+    };
+
+    this.styles = composeStyles(baseStyles, props.stylesheets);
+  }
+
+  componentDidMount() {
+    if (this.props.markedElement) {
+
+      function actualResizeHandler(component) {
```
rysmith: `nb` Is there a reason this is defined inside of `componentDidMount` as opposed to a method on the component?

coach-marks/src/CoachMark.jsx
```diff
@@ -0,0 +1,116 @@
...
+  static defaultProps = {
+    stylesheets: []
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      top: `${props.fixedTop}px`,
+      left: `${props.fixedLeft}px`,
+      right: `${props.fixedRight}px`,
```
dancingshell: could these result in `undefinedpx`?

coach-marks/src/CoachMark.jsx
dancingshell: Does this work for resizing window too?

Jacky-Lei: The linter warned me but I didn't want the component to store a method that it would only use half the time. Keeping it inside the `if` statement here safes some space or processing time I guess?

coach-marks/src/CoachMark.jsx
Jacky-Lei: Sorry didn't see you commented before I merged! I goes result in `undefinedpx` but doesn't make it into the styling

Jacky-Lei: And to the comment below, yeah it works with resizing window

PR #549, opened 2017-11-13
Table Inline Date fix (part 2)

dancingshell:
### Summary

- Fixed bug where selecting date in December cleared select


table/src/InlineDateSelect.jsx
```diff
@@ -196,11 +196,11 @@ export default class InlineDateSelect extends React.Component {
     month *= 1;
     year *= 1;
 
-    if (month === 12) {
+    if (month === 13) {
```
ben-burlingham: nb `> 12` might be more readable and more safe (catches more cases)

PR #550, opened 2017-11-13
Coach marks v 0.1.1

Jacky-Lei:
Fixes throttling on resizeHandler


coach-marks/README.md
```diff
@@ -0,0 +1,21 @@
+# Coach Marks
+
+Overlay with CoachMarks builder
+
+## usage
```
ben-burlingham: nb `U`sage?

Jacky-Lei: I didn't pull before I made a new branch so the refs are all weird :l

PR #552, opened 2017-11-13
[Navigation Tiers] Error State, Spinner Overlay

brock-noah:
### Summary
- Provides error callback for unsuccessful requests
- Spinner overlays on content, smaller spinner, no label
- Disable search when loading
  - correct location for onComplete callbacks because you can't focus on disabled
- Add forgotten onComplete callback

#### Before
- component would be stuck in loading state

#### After
error particle


navigation-tiers/src/navigationTiers.scss
```diff
@@ -134,3 +139,33 @@ $iconColor: #808080;
...
+  max-height: 0;
+  display: flex;
+  align-items: center;
+  overflow: hidden;
+  padding-left: 24px;
+  padding-right: 24px;
+  font-size: 12px;
+  background-color: $errorBG;
+  color: $errorBG;
+  transition: max-height 0.5s linear, color 0.2s ease-in;
```
brock-noah: I transition color here because the React component says "I have no error message anymore", and that toggles the class telling the CSS to start it's animation, but the error message vanishes, leaving the word "Error" in bold. So I do a quick animation to make it seem like all the text disappears at once. Makes it transition a little nicer. Though hopefully they never see this error banner in the first place.

navigation-tiers/src/PageProvider.jsx
```diff
@@ -15,9 +15,12 @@ function key(arr = []) {
   return uuid(arr.join('<(^o^)>'), BASE);
```
chrisarevir: Is this emoji necessary?

brock-noah: An insurance just encase a situation like this happens
```
['Bobs Burgers', 'Wharf']
['Bobs Burgers,Wharf']
```
I want to join on something a little more unique than `,`

PR #557, opened 2017-11-14
Table - Fixed stickyHeader option

tonynd:
### Summary
- Fix stickyHeader option for table. This prevents the header from sticky to the top of the browser window when scrolling up and down the table.
- `stickyHeader` option now default to `false`.

- A change to how the headers stickiness is handled so broke the stickyHeader option.


table/README.md
```diff
@@ -4,6 +4,10 @@ A table component with checkboxes, sticky head, sticky right panel, and horizont
 
 # Change Log
 
+##### 10.0.6
+- Fixed `stickyHeader` option for table.
```
ben-burlingham: Can kill this line, the next one is descriptive and concise

table/src/Table.jsx
```diff
@@ -76,7 +76,7 @@ export default class extends React.Component {
     showCheckboxes: false,
     showSidebar: false,
     sortState: { direction: SORT_DIRECTION.NONE, attribute: null },
-    stickyHeader: true,
+    stickyHeader: false,
```
dancingshell: this seems like a breaking change/ major version bump. is there a reason you needed to switch this?

table/README.md
```diff
@@ -4,6 +4,10 @@ A table component with checkboxes, sticky head, sticky right panel, and horizont
 
 # Change Log
 
+##### 10.0.6
```
ben-burlingham: Unsticky headers will affect many teams, this is a breaking change so should be Major

table/src/Table.jsx
```diff
@@ -100,10 +100,10 @@ export default class extends React.Component {
 
   componentDidMount() {
     window.addEventListener('table/expandAll', this.onExpandAll);
-    window.addEventListener('scroll', this.onScroll);
+    this.props.stickyHeader && window.addEventListener('scroll', this.onScroll);
```
ben-burlingham: Nice

table/src/Table.jsx
tonynd: Talked to @ben-burlingham about this. Would like to keep the consistency of defaultProps in terms of boolean being `false`.

ben-burlingham:
Also can you indicate _Breaking_ in the README - this is something we're doing to help people during their updates, so they can target fixes faster.  Something like:
> 11.0.0
> - _Breaking_ Changing `stickyHeader` to be `false` by default

PR #560, opened 2017-11-15
Update inline select value passing

chrisarevir:
InlineSelect value passing now plucks display value from data prop.

While this is actually a breaking change, it was made backwards compatible, but with a warning to let people know how to upgrade once that API has been deprecated.


table/src/InlineSelect.jsx
```diff
@@ -29,18 +29,31 @@ export default class InlineSelect extends React.Component {
...
+    value: null
   };
 
   constructor(props) {
     super(props);
 
+    const value = props.data.reduce(
+      (acc, item) => (item.key === props.value ? item.value : acc),
+      null
+    );
```
dancingshell: nb: this loops through all the data, even after finding the match. might be more performant to break out early, perhaps use `Array.find`?

table/src/InlineSelect.jsx
```diff
@@ -52,7 +65,12 @@ export default class InlineSelect extends React.Component {
 
   componentWillReceiveProps(nextProps) {
     if (this.props.value !== nextProps.value) {
-      this.setState({ value: nextProps.value });
+      const value = this.props.data.reduce(
+        (acc, item) => (item.key === nextProps.value ? item.value : acc),
+        null
+      );
+
+      this.setState({ value });
```
dancingshell: do we need the same fallback here?

table/src/InlineSelect.jsx
```diff
@@ -29,18 +29,31 @@ export default class InlineSelect extends React.Component {
...
+      );
+    }
+
     this.styles = composeStyles(baseStyles, [...props.stylesheets]);
     this.table = null;
     this.state = {
       expanded: false,
       filter: null,
-      value: props.value,
+      value: value || props.value,
```
dancingshell: nice 👍

PR #566, opened 2017-11-15
fixed bug with user select none

RyanFuentes:
### Summary
Fixed bug with global css selector `user-select: none` that prevents IE Edge textboxes from being unselectable and having a locked cursor.


nickwmurphy:
Think we can close this. Just fixed in another PR #564

ben-burlingham:
Lol about three of us all did this fix at the same time :) I think Nick's got in there first...recommend closing this PR unless there's something additional.

Just FYI I've overrode the style in my Client for now until my Table updates catch up.

PR #569, opened 2017-11-16
Breadcrumb 0.2.0

dancingshell:
### Summary
Breaking changes to make breadcrumb easier to use
- now crumbs are accepted as props


brock-noah:
Why was this change made? Your first implementation is less opinionated about how the dev structures their data. It can also create the second implementation.

mcclayton:
I agree with @brock-noah. The previous approach was more flexible, less opinionated, and a superset of these changes. Curious also about how this version is "easier" to use?

PR #570, opened 2017-11-16
Tiered select ux updates

ben-burlingham:
### Summary
Matching UX mocks.

Will be a major bump.

API mostly the same with some naming updates.


tiered-select/src/TieredSelect.scss
```diff
@@ -1,259 +1,284 @@
...
     outline: 0;
   }
 
-  &.emptyItemsContainer {
-    align-items: center;
-    background: $xLightGray;
-    justify-content: center;
-    padding: 16px;
-    user-select: none;
+  // .currentParent {
```
piercefrawley: do we want to keep this block?

ben-burlingham: Nope, removing now.

PR #571, opened 2017-11-16
Tooltip Particle

nickwmurphy:
### Summary
- A lightweight tooltip component.
- Import once, and add `data-title` (string) and `data-placement` (one of 'top', 'bottom', 'left', or 'right' --- 'top' is default) to any element on the page.

### Screenshots
tooltip

### Future Work
- More placement options
- Better styling
- Add a `delayTime` prop


tooltip/src/Tooltip.jsx
```diff
@@ -0,0 +1,130 @@
...
+    this.findElements();
+  }
+
+  componentWillUnmount() {
+    this.unbindEvents();
+    window.removeEventListener('scroll', this.hideTooltip);
+  }
+
+  findElements = () => {
+    const elements = [...document.querySelectorAll('[data-title]')];
```
mcclayton: Hey @nickwmurphy,
Thanks for working on this!
I have a few concerns with the way this is placing tooltips on all elements with `data-title` data attributes. 
This is not really a Tooltip particle component, it is more akin to a `TooltipPagePlacer`.
What are your thoughts of adding tooltips in a way where we can make this a more reusable component with better control with semantics such as
```
<Tooltip placement='top' content='I am a tooltip'>
    <div>I am a div with a tooltip<div>
</Tooltip>
```

I assume you looked at https://github.com/wwayne/react-tooltip for inspiration however I think the semantics of solutions like:
https://github.com/react-component/tooltip
and
https://github.com/mcumpl/react-lightweight-tooltip
are significantly more flexible, reusable, and don't muddle with the contents of the entire DOM.

Thoughts?

nickwmurphy: @mcclayton thanks for the feedback! I will take a look at these other implementation styles and gain a better understanding of the differences.

dancingshell: @nickwmurphy I was just wishing we had a tooltip component. This would be a great use case for react 16 portals, and if you want to explore going in that direction I am happy to help.

tooltip/src/Tooltip.jsx
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  showTooltip = (evt) => {
+    const element = evt.target;
+
+    this.setState({
+      placement: element.dataset.placement || 'top',
+      title: element.dataset.title
+    }, () => {
+      this.positionTooltip(element);
```
dancingshell: I would avoid using setState callbacks, as they behave differently and can cause some weirdness when implemented in react 15 clients

tooltip/src/Tooltip.jsx
ben-burlingham: Agreed that `data-title` is a bit heavy handed, might suggest `data-tooltip`.

@mcclayton how would the wrapping syntax get applied to elements that aren't in React?

mcclayton: @ben-burlingham 
Using the wrapping syntax to denote an element that should receive a tooltip is no different than denoting an element via `data-tooltip` attribute other than it is more explicit, does not muddle a namespace, allows for better control, and provides flexibility in being able to provide different styles/props to individual tooltips.
Does that make sense?

brock-noah:
If the goal of this is to be used on non react elements, imo this should be vanilla JS.

If it's not being used on react elements, why do you need to instantiate JSX to effect items outside of your client?

tooltip/src/Tooltip.jsx
ben-burlingham: I understand the wrapping syntax and that's all generally true.

My question was, what if we need to apply a tooltip outside of the React component?  If it's an ERB-based markup, then it can't be wrapped in a React handler.

ben-burlingham:
First paragraph - good point, true.

Second - there's cases in our current platform where react clients are mixed together with ERB rendered markup (since it's harder to do full rebuilds rather than incremental expansion of the client).

ben-burlingham:
Ultimately, it's Nick's implementation, and his decision which way to take it.  

If somebody disagrees with that implementation, that's ok, they're welcome to build their own and use that.  Particles are aggressively isolated and opt-in
for this very purpose. 

If the issues that are outlined here end up popping up, it's easy to make updates. :)

brock-noah:
The second paragraph is direct to my first one. If the goal is to target non react elements, a client like Jobsite which uses Vue could use this out of the box if it isn't JSX/react. If you're using JSX to invoke a stateful function, it could just be JS.

brock-noah:
Above are non-blocking.

I think data-particle-tooltip is the best name because data-tooltip is still really general and existing JS or jQuery might latch on.

nickwmurphy:
Final decision: I am going to ship this implementation as version v0.1, and @dancingshell has an implementation using the wrapping style and portals that will ship as v0.2 or v1.0.

ben-burlingham:
Great! Thanks everyone.

ben-burlingham:
Great! Thanks everyone.

Maybe it would help in the future to specify "non blocking", so we can still have these valuable conversations while maintaining developer velocity.

PR #572, opened 2017-11-16
Select Version 9.0.1

piercefrawley:
### Summary
- Compose styles when component receives new stylesheets


select/src/Select.jsx
```diff
@@ -113,6 +113,12 @@ export default class Select extends React.Component {
     };
   }
 
+  componentWillReceiveProps({ styleVariants, stylesheets }) {
+    console.log('here');
```
ben-burlingham: oops

PR #575, opened 2017-11-16
Inline Date Select Clearable Prop

nickwmurphy:
### Summary
- `unclearable` prop for `InlineDateSelect` added (false by default)


table/sandbox/columnDefinitions/ColF.jsx
```diff
@@ -32,6 +32,7 @@ const ColF = {
     const dateToRender = Math.round(Math.random()) ? { year, month, day } : null;
     return (
       <InlineDateSelect
+        unclearable
```
ben-burlingham: `clearable`...?

chrisarevir: Resolved IRL!

ben-burlingham:
Just talked IRL - `unclearable` was to prevent a breaking change.

This is a great candidate for @dancingshell's "breaking change consolidation" plan, flipping it to the more consistent `clearable` can be added to that list.

PR #577, opened 2017-11-17
Tooltip V0.2 with portals

dancingshell:
### Summary

* new API with portals!
* tooltips can now be controlled (not hoverable, show prop passed in by parent) allowing it to stay open

### Screenshots
awesomescreenshot-2017-11-17t19-11-32-172z


tooltip/src/Tooltip.jsx
```diff
@@ -1,130 +1,126 @@
...
-  }
+    controlled: false,
+    placement: 'top',
+    show: false,
+    title: ''
+  };
+
+  static propTypes = {
+    controlled: PropTypes.bool,
+    placement: PropTypes.oneOf(['top', 'bottom', 'left', 'right']).isRequired,
```
ben-burlingham: nb Could these be expressed as an enum, and exported?

tooltip/src/Tooltip.jsx
```diff
@@ -1,130 +1,126 @@
...
+    show: PropTypes.bool,
+    title: PropTypes.string.isRequired
```
mcclayton: `nb` I'd consider generalizing this to something like `content`. This gives us the flexibility to turn this into `Proptypes.node` in the future without needing a breaking change and makes it a little more general.

tooltip/src/Tooltip.jsx
```diff
@@ -1,130 +1,126 @@
...
-import composeStyles from '../../shared/stylesheetComposer';
 
-export default class Tooltip extends React.Component {
-  static propTypes = {
-    stylesheets: PropTypes.arrayOf(PropTypes.shape())
-  }
+import TooltipPortal from './TooltipPortal';
+import styles from './Tooltip.scss';
 
+const ADJUSTMENT = 8;
```
mcclayton: Is this the padding between the tooltip pointer and the component it's wrapping?
Might be nice to add a comment if so.

tooltip/src/Tooltip.jsx
```diff
@@ -1,130 +1,126 @@
...
+    if (this.state.show && prevState.show === false) {
+      this.position();
+    }
   }
 
-  hideTooltip = () => {
-    this.setState({ title: '' });
-  }
+  onMouseEnter = () => {
+    if (!this.props.controlled) {
```
mcclayton: I love the idea of being able to control the tooltip from a parent or have the state self-managed.
In the future I could easily see the self-managed tooltip state being managed by a HoC. This seems to be equally great though :)

tooltip/src/Tooltip.jsx
```diff
@@ -1,130 +1,126 @@
...
-        break;
+  onMouseLeave = () => {
+    if (!this.props.controlled) {
+      this.setState({ show: false });
+    }
+  };
+
+  position() {
+    if (this.parentElement && this.tooltip) {
+      const {
```
mcclayton: `nb`: Not blocking - just pointing out that not all of these properties are available to older browsers:
Here's an example of what IE 8 vs. 9 returns -> https://stackoverflow.com/a/17686282
Not blocking though.

mcclayton:
You'll likely want a throttled window resize handler to help reposition the tooltip on window resize. This will be evident for controlled tooltips.

mcclayton:
Final thing I'd call out (all of my comments are non-blocking)
is that the styling is slightly different from what @nickwmurphy had.
screen shot 2017-11-17 at 11 55 20 am
screen shot 2017-11-17 at 11 56 00 am

This may be intentional, but calling out here in case it wasn't.

tooltip/src/Tooltip.jsx
dancingshell: Removed and re-implemented with css

dancingshell:
@mcclayton @nickwmurphy @ben-burlingham 
Updates
- Added `popover` prop that applies popover styling
- Export position options
- Controlled tooltips now reposition on page resize
- Implemented old adjusted position var with css

tooltip/src/Tooltip.jsx
```diff
@@ -63,7 +85,17 @@ class TooltipWrapper extends React.Component {
...
+    if (!this.throttleTimeout) {
+      this.throttleTimeout = setTimeout(() => {
+        this.throttleTimeout = null;
+        this.position();
+      }, 66);
+    }
+  }
+
   position() {
+    console.log('trying')
```
mcclayton: ^ console

dancingshell: dang 🙃

dancingshell:
also here is the popover styles 
popover

PR #581, opened 2017-11-20
Tooltip: Style Variant "Fixed"

nickwmurphy:
### Summary
- New `styleVariants` prop with `'fixed'` as an option
- Tooltip stays open when hovering over the tooltip element itself
- `delay` prop is now `delayShow`
- Sandbox cleanup


tooltip/src/css-overrides/fixedStyleOverrides.scss
```diff
@@ -0,0 +1,3 @@
+.tooltipContent {
+  position: fixed;
+}
```
dancingshell: yay style variants!

nb: I think in the other particles we had named the file after the variant `css-overrides/fixed.scss`

tooltip/src/Tooltip.jsx
```diff
@@ -38,13 +48,18 @@ export default class TooltipWrapper extends React.Component {
...
+    styleVariants: PropTypes.arrayOf(PropTypes.string)
   };
 
   constructor(props) {
     super(props);
 
-    this.styles = composeStyles(baseStyles, [...props.stylesheets]);
+    const { stylesheets, styleVariants } = props;
+
+    this.styles = composeStyles(baseStyles,
```
dancingshell: also add this to `componentWillReceiveProps`?

nickwmurphy: :+1: done

tooltip/src/Tooltip.jsx
```diff
@@ -143,6 +170,10 @@ export default class TooltipWrapper extends React.Component {
     const { children, content, placement, popover } = this.props;
     const { show } = this.state;
 
+    if (!content) {
+      return null;
+    }
```
dancingshell: only hide tooltip if no content, still render trigger

tooltip/src/Tooltip.scss
```diff
@@ -7,6 +7,7 @@ $tooltipBackgroundColor: rgba(19, 31, 51, 0.9);
   box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
   border-radius: 2px;
   color: $white;
+  max-width: 400px;
```
dancingshell: max width should be 200px for tooltip, 250px for popover

PR #582, opened 2017-11-21
Tiered Select: Key Listeners

nickwmurphy:
### Summary
- Added support for arrow up, down, left, and right and enter keys
- Removing `itemsTitle` because there the data model does not support it in the intended fashion

### Screenshots
tiered-select-key-listeners


tiered-select/src/TieredSelect.scss
```diff
@@ -213,10 +214,9 @@ $totalH: 328px;
   cursor: pointer;
   font-size: 13px;
   line-height: 20px;
-  margin: 8px 0;
-  padding: 0 16px;
+  padding: 8px 16px;
 
-  &:hover {
+  &.focusedItem {
```
chrisarevir: Seems like this is a consistent property of focusedItem, does it need to be in multiple places?

PR #583, opened 2017-11-21
ControlledSelect

nickwmurphy:
### Summary
- To allow for flexibility in data filtering, this branch exports `ControlledSelect` by default. `ControlledSelect` applies a basic character-matching filter method to `data`. (that method is exported as `defaultSearchFn`).
- `Select` is also exported, but with no internal support for filtering. The `onSearch` prop still allows a developer to extract the user's search term and perform custom filtering on `data`.
- Add support for `titleValue` to be a node as well as a string (for example, a client could pass tokens into the `titleValue` area).
- Remove `renderGroup` prop (`renderItem` remains)
- Remove `shouldFilter` prop
- Added `value` prop to `Select` for the search input

Below is a trimmed version of `ControlledSelect`. It essentially just wraps `Select` and controls the `onSearch`, `data`, and `value` props in a way that is considered most common.
```
export function defaultSearchFn({ data, value = '', valueField = 'value' }) { ... }

export default class ControlledSelect extends React.Component {
  constructor(props) {
    super(props);

    this.state = {
      data: props.data,
      value: ''
    };
  }

  componentWillReceiveProps({ data }) {
    this.setState({ data });
  }

  onSearch = ({ value, scope }) => {
    let { data } = this.props;
    const { valueField } = this.props;

    data = defaultSearchFn({ data, value, valueField });

    this.setState({ data, value });
    this.props.onSearch({ value, scope });
  }

  render() {
    return (
      <Select {...{ props }} />
    );
  }
}
```

### Questions
1. Where should `defaultSearchFn` be exported from? (answer: `ControlledSelect`)
2. Where should `SELECT_ANALYTICS` be exported from? (answer: `Select`)

### To Do
- When using `ControlledSelect` and calling `setState()` in the `onSearch` prop, filtering does not work as expected. Need to fully understand why and fix or document. (answer: @dancingshell fixed this)
- Add testing coverage for `ControlledSelect`

### Other
- Removed all remnants of `storybook`
- Removed `hasFocus` as part of the internal state of `Dropdown` (wasn't necessary)


select/README.md
```diff
@@ -4,6 +4,10 @@ A select with support for group or individual selections.
 
 # Change Log
 
+##### 9.1.0
+- Add support for `titleValue` to be a node
+- Remove `shouldFilter` prop (was doing the same as `searchable`)
```
piercefrawley: nice!

ben-burlingham:
1 - directory bulk add picker has different renderers, so if there's a way to support that perhaps there's no need for both
2 - `searchable` seems best, I think you made a good call there

select/sandbox/sandbox.jsx
```diff
@@ -1,5 +1,120 @@
...
+          onSearch={this.onSearch}
+          onSelect={this.onSelect}
+          onSelectAll={this.onSelectAll}
+          renderEmpty={this.renderEmpty}
+          renderGroup={this.renderGroup}
+          renderItem={this.renderItem}
+          searchable
+          slotBottom={slotBottom}
+          slotTitle={slotTitle}
+          style={{ width: '250px' }}
```
ben-burlingham: hmm

nickwmurphy: This is from a while ago. Each `Select` was getting its own stylesheet to set its width in Procore, so this is a way to avoid having a stylesheet for every sized `Select`. We should explore other methods maybe?

ben-burlingham: Sure, no opinions here, just suspicious of inline styles

select/src/Dropdown.jsx
```diff
@@ -247,55 +202,66 @@ export default class Dropdown extends React.Component {
...
-          onClick={() => { onItemClick(item); }}
-          onMouseMove={() => this.onHoverItem(i)}
-        >
-          {renderGroup ? renderGroup(item, i) : item[valueField]}
-        </div>
-      )) : null;
-
-    const selectAll = onSelectAll && data.length
-      ? (<button className={styles.controlsButton} onClick={this.onSelectAll} tabIndex={-1} type='button'>
+    const { focusedItem } = this.state;
```
dancingshell: NB: this is unused in render

select/src/ControlledSelect.jsx
```diff
@@ -0,0 +1,175 @@
...
+
+    if (match && item.children) {
+      collect(item.children, included);
+    }
+
+    return match || included.find(i => i === item.key);
+  });
+}
+
+export default class ControlledSelect extends React.Component {
```
dancingshell: are we going with `ControlledSelect` for the name?

nickwmurphy: @ben-burlingham @piercefrawley ?

select/src/ControlledSelect.jsx
```diff
@@ -0,0 +1,175 @@
...
+    super(props);
+
+    this.state = {
+      data: props.data,
+      value: ''
+    };
+  }
+
+  componentWillReceiveProps({ data }) {
+    this.setState({ data });
```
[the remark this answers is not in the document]
nickwmurphy: 👍 done

select/src/ControlledSelect.jsx
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  onSearch = ({ value, scope }) => {
+    let { data } = this.props;
+    const { valueField } = this.props;
+
+    data = defaultSearchFn({ data, value, valueField });
```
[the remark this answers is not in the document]
nickwmurphy: 👍 done

select/src/ControlledSelect.jsx
```diff
@@ -0,0 +1,175 @@
...
+
+    this.setState({ data, value });
+    this.props.onSearch({ value, scope });
+  }
+
+  render() {
+    const { data, value } = this.state;
+
+    return (
+      <Select
```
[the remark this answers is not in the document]
nickwmurphy: 👍 done

select/src/Dropdown.jsx
```diff
@@ -39,72 +16,50 @@ export default class Dropdown extends React.Component {
...
-    onSearch: () => {},
     onSelectAll: null,
     renderEmpty: null,
-    renderGroup: null,
     renderItem: null,
-    searchable: false,
-    shouldFilter: false,
-    slotBottom: null,
-    valueField: 'value'
+    slotBottom: null
```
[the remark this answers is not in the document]
nickwmurphy: Is required by `Dropdown`, default is in `Select`

select/src/Dropdown.jsx
nickwmurphy: It is! Promise.

select/src/Dropdown.jsx
```diff
@@ -157,7 +112,7 @@ export default class Dropdown extends React.Component {
 
   onSearch = (evt) => {
     const { value } = evt.target;
-    this.setState({ focusedItem: 0, value }, () => {
+    this.setState({ focusedItem: 0 }, () => {
```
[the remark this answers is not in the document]
nickwmurphy: 👍 done

PR #586, opened 2017-11-26
Modal Positioning

nickwmurphy:
### Summary
- Prevent `documentElement` from scrolling while open (this is the desired UX behavior)
- Use position: fixed instead of absolute (this allows the overlay to cover the entire body and not just the component's parent
- Linting (changed a few small things to match other particles)
- Updated padding to better match mocks

### Questions
1. Should we set a really high `z-index` on this particle or just leave it up to the client to pass one if needed? Fixed positioning should place the overlay and modal on top of most elements by default.


modal/src/Modal.scss
```diff
@@ -5,33 +5,34 @@
   height: 100vh;
   justify-content: center;
   left: 0;
-  position: absolute;
+  position: fixed;
   top: 0;
   transition: opacity 0.15s ease-out;
   width: 100%;
+  z-index: 10000;
```
nickwmurphy: Yes or no? In Procore, the sidebar uses some fixed positioned elements with z-indices, so in pretty much every case, this will need to be passed in if not included here in the base particle.

ben-burlingham: This seems reasonable, one of the few exceptions :)

modal/src/Modal.scss
```diff
@@ -5,33 +5,34 @@
...
 
 .header {
   border-radius: 4px 4px 0 0;
-  padding: 32px 48px;
+  padding: 32px 88px 24px 48px;
 }
 
 .body {
-  padding: 32px 48px;
+  padding: 24px 48px;
```
ben-burlingham: You sure these are right? I think I was following the InVision prototypes fairly closely 🤔

modal/src/Modal.scss
```diff
@@ -5,33 +5,34 @@
...
+  box-shadow: 0 3px 6px 0 rgba(0, 0, 0, 0.23), 0 3px 6px 0 rgba(0, 0, 0, 0.16);
   line-height: 24px;
   min-width: 520px;
   position: relative;
 }
 
 .header {
   border-radius: 4px 4px 0 0;
-  padding: 32px 48px;
+  padding: 32px 88px 24px 48px;
```
ben-burlingham: Side note mostly for remembering purposes - I'm not sure that the close button is available if the header content is not passed.  This should be fixed at some point.

modal/src/Modal.scss
nickwmurphy: Yup! It was like 95% correct, just a few touch ups.

modal/src/Modal.scss
nickwmurphy: Just checked, and it is still available when header is not passed 😄.

brock-noah:
Excited to see this! I actually made a component this weekend for anything that needs body lock behavior. 

I did the same thing to the body. Though on Windows computers, if the body has a scroll bar, the scroll bar disappears and the page body will shift a little. It seems like it would take a lot of work to handle this, so if UX is fine with this behavior or an eventual follow up, I'll make a PR for the plain component.

ben-burlingham:
> Though on Windows computers, if the body has a scroll bar, the scroll bar disappears and the page body will shift a little

Nice catch - just out of curiousity, how are you testing on Windows machines?

modal/src/Modal.scss
ben-burlingham: Awww yiss

brock-noah:
Run sandbox locally.
Use [ngrok](https://ngrok.com/).
Go to the URL from a Window machine.

PR #587, opened 2017-11-27
Km/calendar fixes

kedarmoghe:
fixing a bug with setting range (style related)


calendar/src/CalendarDay.jsx
```diff
@@ -28,23 +28,26 @@ const isBetweenRange = ({ committedDates, date, hoverDate, step }) => {
   const endDate = getEndDate(committedDates);
   const validHoveredDate = isValidHoveredDate(hoverDate);
 
+  let adsf = false;
```
ben-burlingham: Oops, looks like forgot to update to a more descriptive name here

PR #588, opened 2017-11-27
Text Input v3

nickwmurphy:
### Summary
- Add `autoFocus` prop
- _Breaking_: Add `clearable` prop
- Style updates


text-input/README.md
```diff
@@ -3,6 +3,12 @@ Text Input Component
 
 # Change Log
 
+##### 2.1.0
+- Add `autoFocus` prop
+- Add `clearable` prop
+- Add `type` prop
```
ben-burlingham: What's the use case for the `type` prop? Seems like a text input will always be `text`

text-input/src/TextInput.jsx
```diff
@@ -95,62 +110,77 @@ export default class TextInput extends React.Component {
...
+    const spinner = loading && (
+      <div className={this.styles.iconHolder}>
+        <span className={this.styles.spinner} />
+      </div>
+    );
+
+    return (
+      <div>
+        {clearButton}
+        {spinner}
```
dancingshell: NB: `{clearButton || spinner}`? I don't think they can happen at the same time. The logic in here could probably be cleaned up to make it more readable, but that can happen later

PR #592, opened 2017-11-28
Table - Added hover options for row and sidebar

tonynd:
### Summary
- Added `onRowHover` option. Ability to kick off an action when hovering over a row.
- Added `onSidebarCellHover` option. Ability to kick off an action when hovering over a sidebar cell.

- Both options returns the associated `rowData`.


table/README.md
```diff
@@ -4,6 +4,10 @@ A table component with checkboxes, sticky head, sticky right panel, and horizont
 
 # Change Log
 
+##### 10.2.2
```
dancingshell: New features = `10.3.0` patch

table/src/BodyRow.jsx
```diff
@@ -62,6 +63,8 @@ const BodyRow = ({
     className={styles.tableRow}
     style={{ minHeight: `${rowHeight}px` }}
     onClick={() => onClick(rowData)}
+    onMouseOver={() => onHover(rowData, 'enter')}
+    onMouseOut={() => onHover(rowData, 'leave')}
```
dancingshell: NB: I might vote for switching params interface to be `(type, data)`, but i don't feel that strongly either way

ben-burlingham:
Talked with @tonynd about this, it's a pretty complicated issue and will trigger a re-shuffle of the markup and the way the sidebar works.  I'll get on that soon.

For now, we can ship this solve to unblock @tonynd, with the caveat that it will not be the permanent fix.

PR #595, opened 2017-11-28
next.md attempt to consolidate breaking changes

dancingshell:
### Summary

This PR is a result of conversations regarding major version bumps for particles, in an attempt to consolidate breaking changes into one major release if it is possible.

Need a breaking change? Ask these questions:
1) Can I accomplish this in a different way that will make it backwards compatible?
    * **Yes?** -> Make non breaking change. If further breaking changes are still desired (for example to clean up prop names), log the desired change in the `NEXT.md` with your github handle so future implementors can contact you with any questions.
   * **No?** -> Prepare for major release by referencing the component's `NEXT.md` file to see if other changes were staged to go out in the next major release. You should take the lead to coordinate with the participants requesting these changes to arrange the release.
2) Keep `NEXT.md` up to date by checking off the requested changes once they are incorporated

This was a basic attempt to organize around this idea. I expect this to evolve with time as we see the flow in action. Feel free to leave feedback!


ben-burlingham:
This looks awesome - can we simplify the markdown a bit?  I'd suggest 
- Removing `isComplete` since if it's in the list, it's not complete
- Removing the `ul` `li` part (unless screenshot?) since it just adds to maintenance

So something that would pretty much just be a list:
- @theperson MM-DD-YYYY describe the thing that should be done with its notes
- @theperson MM-DD-YYYY describe the thing that should be done with its notes
- @theperson MM-DD-YYYY describe the thing that should be done with its notes

dancingshell:
@ben-burlingham here is a screenshot. My thoughts on `Is Complete` is that we would have a kind of running record, but I guess once it ii incorporated it would then be in the README. The gross `ul` `li` syntax was the only way to render checkboxes within a table. 

I'm open to getting rid of the table entirely though, if it's too confusing/clunky

screen shot 2017-11-28 at 2 45 13 pm

ben-burlingham:
Oh I see yeah that's pretty.

Well, gut reaction is yeah the clunkiness is a thing and might discourage people from using it.  Also, it might make it more difficult to check if a feature is already on the list.  

Doesn't matter too much, just personally feel it could be lightened up quite a bit.  Thanks for starting up this pattern!

table/NEXT.md
```diff
@@ -0,0 +1,6 @@
+### Next Major Version Should Include
+
+@chrisarevir `unclearable` -> `clearable` for `InlineDateSelect`
```
dancingshell: can we make these bullet checkboxes now that they aren't in the table?

ben-burlingham: How about just bullets, since once they're complete they'll be removed rather than checked?

PR #606, opened 2017-11-30
Table: Exporting TableWithStatefulCheckboxes decorator.

ben-burlingham:
### Summary
Thanks @RyanFuentes, we're making this dream a reality.

`<Table>` is dumb as a post, if you want checkbox management use the `<TableWithStatefulCheckboxes>` decorator.


RyanFuentes:
When all of a groups children are selected does the group become selected? I think that logic was left out of the sandbox code because it was a bit complex.

ben-burlingham:
Ahh good catch, I'll check that out.

RyanFuentes:
Same goes for the ALL checkbox

table/src/TableWithStatefulCheckboxes.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+    const targetRowIds = this.props.data
+      .find(({ key }) => key.key === groupId)
+      .data
+      .map(r => r[this.props.idKey]);
+
+    if (value) {
+      this.setState({
+        checkedGroups: this.state.checkedGroups.concat(groupId),
+        checkedRows: this.state.checkedRows.concat(targetRowIds)
```
RyanFuentes: Not sure if this is a problem, but lets say there's a scenario where a group contains ids 1,2,3,4 and 5. If 1, 2, and 3 are all checked and then you check the group, will the checkedRows state have duplicates for 1, 2, and 3?

table/src/TableWithStatefulCheckboxes.jsx
```diff
@@ -0,0 +1,69 @@
...
+      });
+    }
+  }
+
+  onCheckAll = (value) => {
+    if (value) {
+      this.setState({
+        checkedGroups: this.props.data.map(({ key }) => key.key),
+        checkedRows: this.props.data
+          .reduce((acc, { data }) => acc.concat(data.map(d => d[this.props.idKey])), [])
```
RyanFuentes: Same question here regarding duplicates. It probably doesn't matter functionally if there are multiple identical ids stored in these arrays but it might be cleaner to convert the array to a set data type, add the ids, and then convert back to an array, or to only add the ids that don't already exist.

ben-burlingham:
Nice catch @RyanFuentes.  

It is not an issue on `onCheckAll` since all the checkboxes are assembled from scratch, but you're right on with the`onCheckGroup` issue.  I've made sure it's a `Set`.

PR #613, opened 2017-12-01
add qa tags

evandapolarbear:
### Summary
- Added a couple 'data-qaTag' to enable automated testing


date-select/src/DateSelect.jsx
```diff
@@ -419,6 +419,7 @@ export default class DateSelect extends React.Component {
       value.day || value.month || value.year ? (
         <button
           className={this.styles.clearButton}
+          data-qaTag='particlesDateSelectClearButton'
```
ben-burlingham: Can we also export an enum with these strings, similar to analytics?

evandapolarbear: 👍

PR #614, opened 2017-12-01
Select qa

evandapolarbear:
### Summary
- Added some qa tags


select/src/Select.jsx
```diff
@@ -207,6 +213,7 @@ export default class Select extends React.Component {
       ? (
         <button
           className={this.styles.caret}
+          data-qaTag='particlesSelectCaret'
```
ben-burlingham: enum

select/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@particles/select",
-  "version": "9.0.2",
+  "version": "9.0.3",
```
ben-burlingham: minor? (new functionality)

PR #615, opened 2017-12-01
Removed setState callback from InlineSelect

dancingshell:
### Summary

- onSelect was not firing in React 15 clients, this fixes that 🙃


ben-burlingham:
...pending merge conflicts

PR #617, opened 2017-12-02
[Breadcrumb] Support for Custom Elements on Crumb

brock-noah:
### Issue
Support for React Router.
It would be nice to use the `Link` component with the `to` prop rather than something like
```
<Crumb onClick={e => routerProps.history.push('to')}>
```
(`Crumb` currently does not support `onClick` either for the above to work.)

Ideal:
```
<Crumb to="/location" element={Link} title="Location" />
```

### Change Log
- add `Crumb.props.element` for non active elements

React Router Link [supports](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L76) this because it spreads the props and returns an `a`

### Discussion Topics:
- Is there a naming convention for this feature? Should the prop be called `element`, `component`, `tag`, etc.


ben-burlingham:
Not blocking - and pardon my ignorance, don't want to get involved, but can't help but weigh in here - it seems to me that this is a great example of where the declarative pattern is going a bit too far.

By obfuscating the markup in `Crumb`, we've now:
0. Made it more difficult to integrate with other applications (react router, IIRC)
1. Made one more maintenance task by looking for what element `Crumb` is. It's a `<div>`, which means we've
2. Lost support for hrefs (although the "href" property suggests a link is there?)
3. ...which we solve by wrapping an `<a>` with a set of more obfuscated props: `<a {...props}>{props.children}</a>`

So we've now created a problem and solved it with another layer of obfuscation.  I'm confused as to why the component simply wouldn't support passing it a data set of `{ href, label }`, having `.crumb` and `.active` classes, and call it a day.

Further, a novice initiate (sometimes referred to as "interns", or "me") may have trouble (or not want to spend the time) decoding this, and be inclined to not fix anything for fear of breaking something else.  Anyways, sorry to intrude.  👋

breadcrumb/src/Crumb.jsx
```diff
@@ -2,23 +2,25 @@ import React from 'react';
...
-export default function Crumb({ title, href, active, styles }) {
+export default function Crumb({ element: Element, title, href, active, styles, ...rest }) {
   const classes = cx(styles.crumb, { [styles.active]: active });
   if (active) {
-    return (<span className={classes}>{title}</span>);
+    return (<span {...rest} className={classes}>{title}</span>);
   }
 
-  return (<a href={href} className={classes}>{title}</a>);
+  return (<Element {...rest} href={href} className={classes}>{title}</Element>);
```
chrisarevir: It looks very flexible here, but at the same time I wonder what the use of this is now? Other than "being declarative", why not just use an `<a  {...} />` instead of a `<Crumb element={<a />} ... />`, which seems more `imperative` than `declarative`?

dancingshell: I also thought about using crumb as more or a styling wrapper and rending its children. I agree its almost superfluous at that point though 🤷‍♀️ 

```js
<Crumb><a href={href}>{title}</a></Crumb>
<Crumb><Link to={href}>{title}</Link></Crumb>
<Crumb active>{title}</Crumb>
```

brock-noah:
@ben-burlingham 

0. Not sure how it makes integration more difficult.  I added support for integrations by changing a hardcoded item to a variable.
1. `Crumb` is not a div
2. We never lost support for `href`
3. This allows you to pass in custom anchor elements, you don't have to use it

The component cannot use only `{ href, label }` because react-router handles the history navigation, not browser reloads from links.

breadcrumb/src/Crumb.jsx
brock-noah: @chrisarevir 
The issue with just an anchor is applying the style. We could write CSS like `.breadcrumbs a`. Maybe that is better.

We are not using `element={<a />} />`, it is `element={(propsFromCrumb) => <a />} />`. This allows `Crumb` to deal with providing custom/specific props (like class name) to a custom component that doesn't use an href for navigation.

@dancingshell 
I thought about that too. Similiar issue as I mentioned above, the class needs to be applied to the `a`. When having the wrapper with the class, say we add padding, you are not expanding the actual click area.

If we choose to make crumb just a wrapper like:
```
<Crumb active><a>Label</a></>
```
We would need CSS like `.crumb--active > a` to target anchors.

Maybe that is be preferable? I went for not changing the implementation in this PR. The anchor is taken care for you, but if you need custom anchors for whatever reason, you are free to pass in a component.

PR #618, opened 2017-12-04
Calendar Update

kedarmoghe:
adding disableFutureDates prop and fixed width to calendar


dancingshell:
Could we possibly want to be more flexible in the disabling of dates? Perhaps we want to allow just two weeks of valid dates, or a month. Instead of adding more props for all of these configurations, maybe instead we could have the props be `minDate` and `maxDate` which you would have to adjust in the client based on the date value. Then if you want to disable future dates you can set `maxDate` = `today`

kedarmoghe:
@ben-burlingham gotcha.  In this case we are deferring to moment localization.  Is that fine for this use-case?  That way you don't need to send it i18n for date labels (ie, M = Monday, but in Spanish it is L = Lunes).

ben-burlingham:
@kedarmoghe it's up to you, since you're closest to the ground on the use cases and solutions.

Just mentioning conventions so you're aware they're in place :)

calendar/README.md
```diff
@@ -4,6 +4,10 @@ A calendar component.
 
 # Change Log
 
+##### 3.0.0
```
dancingshell: don't think this requires major version bump. `2.1.0` should do it

PR #619, opened 2017-12-04
Filters calendar onFocus/onBlur edge cases fix

justinmwatts:
### Summary
- Fixing calendar filter onFocus and onBlur edge cases. When a user focuses on an input, then clicks off the calendar, the filter doesn't close unless the user clicks the token. Also, when a user clicks on random spots of the calendar, the calendar closes when it should remain open.


ben-burlingham:
@justinmwatts the `ignoreBlur` is not a great solve, did you find it to be helpful?

PR #622, opened 2017-12-05
Add componentWillUnmount hook

scottwalstead:
### Summary
Some tabs in our project don't have filters so switching tabs causes the whole filters component to unmount

We want to hook into the unmount lifecycle event and pass through the currently saved filters to back them up so we can use them again when the component is remounted


brock-noah:
Will the existing `onSave` callback work?

scottwalstead:
Closing this PR since there is another solution to handle this - we can fetch the filters from the query string in our apps url. Initially thought particles unmount would be cleaner but they are about the same

@brock-noah the onSave callback isn't called when we unmount the component

brock-noah:
My comment was all filter changes go through `onSave`. Nothing has changed since the last time save was called and unmounting.

PR #626, opened 2017-12-05
PhoneInput Component

pswami:
### Summary
- Create PhoneInput Component
   - Used to format phone number using google's `libphonenumber`

### Screenshots
screen shot 2017-12-05 at 10 43 28 am

- Add more countries
- Add more tests
- Replace Flags with SVGs


phone-input/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "url": "https://github.com:procore/particles/"
+  },
+  "author": "Procore",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/procore/particles/issues"
+  },
+  "homepage": "https://github.com/procore/particles/tree/master/phone-input",
+  "dependencies": {
+    "@particles/select": "^10.0.0",
```
ben-burlingham: It's strongly recommended to avoid cross-pollination with particles, to avoid side effects as things change. (`<Select>` has undergone 10 major version changes, for example.)

In other words, if someone changes select, your particle might break unexpectedly.  There's many other reasons as well that I won't go into here (`left-pad`, filesize, maintenance, it's a whole lunch and learn.)

pswami: Talked with Ben offline. I did this for the sake of consistency but it may end up being a bad idea. I can swap this out for `react-select` in a follow up

PR #629, opened 2017-12-06
Select Dropdown is Portaled

dancingshell:
### Summary

In small sections the dropdown content would cause the section to become scrollable when opened. Not anymore with portals!

### Notes
I did remove width 100% from the dropdown css, but i don't think this is breaking because this width used to be relative to the parent width of the select, and now I am setting that dynamically via the portal positioning so the results will be the same. Open to discussion about making it a major version bump if ya'll feel differently


dancingshell:
Closing this, too far behind. May come back to it if there is more of a need

PR #638, opened 2017-12-08
Scoped Scroll

brock-noah:
### Summary
Some JS that only allows scrolling for the container your mouse is over.

## Usage
```
import scopedScroll from '..'

// applies event listener for 'wheel' on element
scopedScroll.add(document.getElementById(..))

// removes event listener on element
scopedScroll.remove(document.getElementById(..))

// the function determining if scroll should happen
scopedScroll.scrollInside(event)
```


ben-burlingham:
Just talked to Brock IRL and the problem trying to be solved here is that when a container scrolls to the bottom, the body then picks up the scrolling action and is scrolled to the bottom as well.

This is default browser behavior. A UX request has asked to override that.

Noting here for posterity that a Particle may not be the best way to attack this, I'm concerned that this will only be used once and then fall into an Indiana Jones archive of forgottenness.  It may be best to solve this on the Client side, rather than the Particles side.

PR #639, opened 2017-12-11
[GLU] Toast Alert

mboperator:
### Summary
Added a Toast that shows and dismisses when `active={true}` is passed

### Screenshots
toastalert


ben-burlingham:
`notification-bar`

mboperator:
Hey @ben-burlingham thanks for checking this out.
I wasn't familiar with `notification-bar` so I spun it up.

The main things here are what I'm implementing is a `Toast` which actually serves a different from the `Banner` paradigm that's being implemented in `notification-bar`

The main differences are:
- a user can never interact with `Toast` (ie it's none dismissable)
- there is only ever a success `Toast`
- `Toasts` are shown at the bottom of the page outside of the flow of other components
- `Banners` can be persistent
- `Banners` are displayed within the flow of the page

Though they look similar (because of the beautiful color palette, no doubt) the use cases are very different. 

`notification bar` conflates a lot of these use cases -- which is cool flexibility wise -- but I'm worried that the variety of usages will lead developers to unintentionally make experiences that are not in line with what UI/UX wants. 

One option moving forward I can think of is keeping `Toast Alert` to address the `Toast` usecase, and modifying `notification bar` to serve the more complex `Banner` usecase.

PR #640, opened 2017-12-11
[Table] Draggable Rows

brock-noah:
### Summary
Allows drag and drop interaction of rows. Provides two new props:
- Boolean `enableDraggableRows`
- Function `onRowDrop(from = { rowIndex, groupIndex }, to = { rowIndex, groupIndex })`

### Screenshots
dnd

### Review Notes/ Questions
- makes no assumptions about dragging between groups, simply provides which groups
- is `enableDraggableRows` too verbose? wanted to show a boolean value and it effects rows

### Next Steps
- ~get SVG for drag handle (this PR)~
- fix bug of not being able to be dropped as last row (maybe not this PR)
- UX decides if entire row or just the handle is draggable (maybe not this PR)
- ~bump and build (this PR)~


table/sandbox/sandbox.jsx
```diff
@@ -110,12 +110,20 @@ class Sandbox extends React.Component {
...
           data={this.state.data}
+          enableDraggableRows
           globalKey='table-sandbox'
           idKey='id'
           onAnalytics={this.onAnalytics}
           onChange={this.onChange}
           onMounted={this.onMounted}
           onRowClick={rowId => console.warn('Row clicked: ', rowId)}
+          onRowDrop={(start, end) => {
+            console.log('from', start, 'to', end);
```
Jacky-Lei: Intentional?

table/src/Table.jsx
```diff
@@ -287,12 +302,45 @@ export default class extends React.Component {
...
+  }
+
   updateStorage = () => {
     const { collapsedIds } = this.state;
 
     sessionStorage.setItem(`${this.props.scope}/COLLAPSED`, JSON.stringify(collapsedIds));
   }
 
+  windowDrop = (e) => {
+    e.preventDefault();
```
Jacky-Lei: Curious what is the default behavior that this all the others are preventing?

ben-burlingham: That's probably required for drag-drop to work properly:

> To change that behavior so that an element becomes a drop zone or is droppable, the element must have both ondragover and ondrop event handler attributes. 

(https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API)

table/sandbox/sandbox.jsx
ben-burlingham: Yep, it's for feedback to the dev in the sandbox environment :+1:

table/src/BodyRow.jsx
```diff
@@ -60,31 +67,80 @@ const BodyRow = ({
...
+      }}
+    />);
+  }
+
+  let ghost = null;
+  if (ghostRowIndex === rowIndex && ghostGroupIndex === groupIndex) {
+    ghost = (<div
+      key='ghost-row'
+      className={styles.ghost}
+      style={{ minHeight: `${rowHeight}px`, minWidth: `${dimensions.tableWidth}px` }}
```
ben-burlingham: nb This could be trouble down the line...concerned about perf and maintenance

ben-burlingham:
You're a beast @brock-noah! Thanks for taking this on

table/src/BodyRow.jsx
brock-noah: Specifically the style prop?

PR #643, opened 2017-12-12
Select inline styleVariant hover state v2

dancingshell:
### Summary

- Adds new orange edit hover styles to inline styleVariant

### Screenshots
screen shot 2017-12-11 at 4 22 47 pm
screen shot 2017-12-11 at 4 22 38 pm


select/sandbox/selectWithDefaultSearchSandbox.jsx
```diff
@@ -97,6 +97,7 @@ export default class SelectWithDefaultSearchSandbox extends React.Component {
           </label>
         </form>
         <SelectWithDefaultSearch
+          styleVariants={['inline']}
```
ben-burlingham: In general, can we start exporting these as an enum, rather than magic strings?  That way we can console log them in the sandbox, reduce typos, etc.

PR #647, opened 2017-12-13
[Nav Tiers] Fix bug when closing the dropdown with no search results

brock-noah:
### Summary
- list does not exist when search results return nothing, apply function directly to element


navigation-tiers/sandbox/sandbox.jsx
```diff
@@ -6,6 +6,7 @@ import OtherTier from './StartOnNthTier';
 render(
   <div style={{ display: 'flex', justifyContent: 'space-around' }}>
     <link href='http://procore-core.global.ssl.fastly.net/3.0.1/core-icons/icons.css' rel='stylesheet' />
+    <style type='text/css'>{'body { height: 125vh; border: 5px dotted dodgerblue; }'}</style>
```
ben-burlingham: `dodgerblue` is a personal favorite of mine

PR #659, opened 2017-12-15
[Table] Internal use of react range

dancingshell:
### Summary

* Fixes bug in some clients: `Cannot read property 'func' of undefined` from `react-range` component


table/src/HorizontalScroller.jsx
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import cx from 'classnames';
-import Range from '@mapbox/react-range';
+import Range from './util/ReactRange';
```
ben-burlingham: Rather than making a new `util` folder for separate components, can we either a) leave it flat in the root directory or b) just inline it into the actual `HorizontalScroller`, since that's what it's doing anyway?

ben-burlingham: Talked to @dancingshell IRL, we'll push as is and update later.

PR #662, opened 2017-12-15
Table: Props listener for stateful checkboxes. Console error fix.

ben-burlingham:
### Summary
Adding a `checkedRows` prop update listener.
Restoring a missing event from `ReactRange` update.
Linting.


table/src/ReactRange.jsx
```diff
@@ -6,31 +6,38 @@ export default class Range extends React.Component {
...
     onMouseMove: () => {}
   };
+
   componentWillReceiveProps(props) {
     this.range.value = props.value;
-  };
-  onRangeChange = () => {
+  }
+
+  onRangeChange = (e) => {
```
dancingshell: my bad

PR #666, opened 2017-12-15
[Table] Fix minor table resize issues in Horizontal Slider

scottwalstead:
### Summary
There was a timing issue for the Directory Table where the width of the table wasn't being measured correctly because it was being measured too early in HorizontalScroller's componentDidMount

Running onResize when the browser triggers a window onload event fixes that timing issue for us

Also fixes onResize pointing to a non existent function (this.reset)


ben-burlingham:
...pending conflicts

nick-morrill:
@scottwalstead I think you may have added me as a reviewer by mistake 😜

scottwalstead:
@nick-morrill yeah. my bad. Added the wrong Nick.

scottwalstead:
@ben-burlingham just resolved them

PR #673, opened 2017-12-19
Notification Bar: Action Buttons

nickwmurphy:
### Summary
- Added props `onAction` (to pass a callback) and `actionTitle` (to pass a string for the action button title)

screen shot 2017-12-18 at 11 12 24 pm


nickwmurphy:
Yes. They work with bars that have no `body` prop and bars that have the `noDismiss` prop.

PR #679, opened 2017-12-20
[Select] Changed inline expanded border color

dancingshell:
### Summary
`inline` styleVariant border now gray while expanded  

##### before
screen shot 2017-12-20 at 2 51 09 pm

##### after
screen shot 2017-12-20 at 2 51 18 pm


ben-burlingham:
Doing more UX updates and the `variables` SCSS is becoming more irrelevant, unfortunately...wonder where that will end up.

PR #690, opened 2017-12-22
Filters: Checking only selected.

ben-burlingham:
### Summary
How many words is a `gif` worth?

### Not the right behavior

"Select Results" affects all checked checkboxes

select results

### The right behavior

"Select Results" only affects checkboxes in search results

filters


chrisarevir:
Assuming a picture is worth 1000 words...

```
const gifWorth = (gif) => gif.frames.length * 1000;
```

PR #693, opened 2017-12-23
[Table] Added polyfill, fixed IE errors

dancingshell:
### Summary

- Fixed IE error on row hover
- Added polyfills to dev mode


chrisarevir:
Why is the polyfill needed? We haven't shipped a polyfill with a component as that is generally the job of the client to include such things. 

Alternatively, If there is a specific polyfill that is needed, it would be better and more light weight to just ship that part.

dancingshell:
@chrisarevir the goal of this was just for sandbox mode, to be able to test the table in IE

table/src/Table.jsx
```diff
@@ -243,7 +243,7 @@ export default class extends React.Component {
 
     this.hoveredRowId = id;
 
-    document.querySelectorAll(`[data-id='${e.currentTarget.dataset.id}']`).forEach((node) => {
+    Array.from(document.querySelectorAll(`[data-id='${e.currentTarget.dataset.id}']`)).forEach((node) => {
```
ben-burlingham: Ahh yes, good ol' `nodeList`

PR #701, opened 2018-01-02
[Breadcrumb] Fix bug with null children (crumbs)

dancingshell:
### Summary

- Allow for conditionally adding crumbs without render error

Alternative solution with props:
`<Crumb href={'http://google.com'} title={'Google'} /> show={false}`

thoughts?


breadcrumb/src/Breadcrumb.jsx
```diff
@@ -10,12 +10,14 @@ export default function Breadcrumb({ stylesheets = [], children }) {
...
   return (
     <div className={styles.container}>
-      {React.Children.map(children, (child, index) => (
-        [
-          React.cloneElement(child, { styles }),
-          index < childrenCount - 1 && chevronIcon
-        ]
-      ))}
+      {React.Children.map(children, (child, index) => {
+        if (child) {
```
ben-burlingham: I might recommend using a `.filter` then `.map` the results, so there aren't empty indexes in the result.

dancingshell: @ben-burlingham `React.Children` doesn't have a filter function, and from what I just read this is to preserve keys for performance reasons. Did you have another idea of how to filter children?

dancingshell: Update: Previous iteration didn't work for conditionally applying the chevrons between the crumbs. Now using `React.Children.toArray` to filter out null children and get more accurate index count. Thanks for bringing attention to this

PR #708, opened 2018-01-03
[Table] 12.0.0 z-index fix and renderContextMenu

ben-burlingham:
### Summary
- _Breaking_ "General to specific" naming convention updates for onGroupCheck (now onCheckGroup), onRowCheck -> onCheckRow, onAllCheck -> onCheckAll
- _Breaking_ "General to specific" naming convention update for onRowClick (now onClickRow)
- Added `renderContextMenu` on body rows
- `.tableContainer` `overflow` `scroll` -> `auto`
- Z-index issue on table config:

### Before
screen shot 2018-01-03 at 9 29 20 am

### After
screen shot 2018-01-03 at 9 47 12 am


ben-burlingham:
I'll do a major bump then. :+1:

Ideally, other particles would not be affected, since they're not composed together for exactly this reason.

table/src/Table.jsx
```diff
@@ -69,15 +70,16 @@ export default class extends React.Component {
...
+    onCheckAll: () => {},
+    onCheckGroup: () => {},
+    onCheckRow: () => {},
+    onClickRow: () => {},
     onRowDrop: (itemFrom, itemTo) => {},
-    onGroupCheck: () => {},
-    onRowCheck: () => {},
-    onRowClick: () => {},
     onSort: () => {},
+    renderContextMenu: null,
```
dancingshell: NB: I see this is to determine if they have contextMenu enabled (or something like that). Are we worried about having all the other func props get a no-op by default, but this gets null?

ben-burlingham: Actually in the next lines, all the other `renderX()` methods are also `null` :)

It could go either way, personally I prefer the `null` since it's an implicit check that the thing should be rendered.  Without it, we'd have to have another prop to say "use a context menu", which could be confusing.

scottwalstead:
Nice change for scroll vs auto
https://stackoverflow.com/questions/6689412/difference-between-html-overflow-auto-and-overflow-scroll

PR #715, opened 2018-01-08
Table: Firefox checkboxes.

ben-burlingham:
### Summary
Empty heckboxes were not rendering as expected on FF. Fixed that and now they're rendering like checkboxes.


dancingshell:
@ben-burlingham any thoughts on merging this patch back to table v11 somehow?

ben-burlingham:
@dancingshell I haven't tried doing that, we'd probably have to submit a couple of PRs and publish an older version and then a newer one...not something we've had to do before.

Good news though, the bump to 12 is not very critical, it was mainly a z-index update.  Can you try to bump to 12 first?

PR #720, opened 2018-01-11
Paginator Key Listeners

nickwmurphy:
### Summary
- Adding support for keyboard navigation (up, down, enter, esc)
- Minor style tweaks


paginator/src/Paginator.jsx
```diff
@@ -109,6 +120,69 @@ export default class Paginator extends React.Component {
...
+    this.setState({ focusedItem: item });
+  }
+
+  scrollDown = () => {
+    if (this.dropdownBodyScroll) {
+      const itemArray = [...this.dropdownBodyScroll.childNodes];
+      const items = itemArray.slice(0, this.state.focusedItem + 1);
+
+      const scrollPosition = items.reduce((acc, item) =>
+        acc + item.clientHeight, 0) - this.dropdownBodyScroll.clientHeight;
```
brock-noah: Does `this.dropdownBodyScroll.childNodes[this.state.focusedItem + 1].offsetTop` work

paginator/src/Paginator.jsx
```diff
@@ -65,8 +65,12 @@ export default class Paginator extends React.Component {
     super(props);
 
     this.styles = composeStyles(baseStyles, [...props.stylesheets]);
+    this.totalPages = Math.ceil(props.totalRecords / props.perPage);
```
brock-noah: Will `totalPages` update if the props change?

paginator/src/Paginator.jsx
```diff
@@ -109,6 +120,69 @@ export default class Paginator extends React.Component {
...
+        break;
+      case 'Enter':
+        this.onChange(focusedItem);
+        this.onHeadExpand(evt);
+        break;
+      default:
+        break;
+    }
+
+    this.setState({ focusedItem });
```
brock-noah: `nb` I honestly don't know if `scrollUp` or `setState` will fire first. I would pass the integer to `scrollUp`/`scrollDown` to remove the state dependency in the function.

paginator/src/Paginator.jsx
```diff
@@ -87,7 +91,14 @@ export default class Paginator extends React.Component {
...
   onHeadExpand = (evt) => {
     evt.stopPropagation();
-    this.setState({ expanded: !this.state.expanded });
+    this.setState({
+      expanded: !this.state.expanded,
+      focusedItem: 1
+    });
+
+    if (!this.state.expanded) {
+      this.dropdownBodyScroll.scrollTop = 0;
```
brock-noah: Is `this.dropdownBodyScroll` guaranteed to be in the DOM?

paginator/src/Paginator.jsx
nickwmurphy: Good point. I will test that and fix if not.

paginator/src/Paginator.jsx
nickwmurphy: Good catch. I'll check for it first.

paginator/src/Paginator.jsx
nickwmurphy: I'll look into this.

paginator/src/Paginator.jsx
nickwmurphy: I'll try it out.

paginator/src/Paginator.jsx
nickwmurphy: I added that same line to `componentWillReceiveProps()`. I think that will take care of it!

paginator/src/Paginator.jsx
nickwmurphy: This was a really great catch @brock-noah! Made me realize I was building with an unknown hack in my scroll methods! All fixed up now.

paginator/src/Paginator.jsx
brock-noah: Is it necessary to store it on the instance? Can it just be a function?

chrisarevir: I think this might be returned in the meta information from the endpoint, so it's possible you dont need to recalculate it here

chrisarevir: I lied, thought this was a client pr

paginator/src/Paginator.jsx
brock-noah: Does this get a similar number? I'd love to remove the spreading, slicing, and reducing.

PR #721, opened 2018-01-11
TableConfig Key Listeners

nickwmurphy:
### Summary
- Added support for keyboard navigation (up, down, enter, esc) to `TableConfig`


nickwmurphy:
It does not. We can certainly try and improve in the future though.

PR #725, opened 2018-01-12
Table minWidth Fix

nickwmurphy:
### Summary
- Fixed bug where the table `minWidth` was incorrect after resizing or showing/hiding columns (#727)
- Only show horizontal scroller when necessary
- Lots of **LINTING**


table/src/Table.jsx
```diff
@@ -188,13 +187,28 @@ export default class extends React.Component {
...
     // Reflow violations here - make sure to trigger sparingly. Ben 171003
-    const tmp = Object.assign({}, dimensions, { tableWidth: this.table.scrollWidth });
+    let tableWidth = columnDefinitions.reduce((acc, def) => {
+      if (!def.hidden) {
+        return acc + def.width;
+      }
+      return acc;
+    }, 0);
+
+    if (enableDraggableRows) {
```
scottwalstead: so our code for the resize does exactly the same thing. Only difference is mine was using refs to measure width instead of just knowing them. Let's discuss this later in our sync up.

table/src/Table.jsx
```diff
@@ -145,6 +143,7 @@ export default class extends React.Component {
     const columnDefinitions = this.columnDefinitionsValidation(nextProps.columnDefinitions);
 
     this.setState({ allCollapsed, allGroupIds, columnDefinitions });
+    this.onResize();
```
scottwalstead: This resolves https://github.com/procore/particles/issues/727, reference it in the PR so it closes the issue when we merge this

nickwmurphy: 👍

table/src/ReactRange.jsx
```diff
@@ -44,7 +45,7 @@ export default class Range extends React.Component {
 
   render() {
     const props = Object.assign({}, this.props, {
-      defaultValue: this.props.value,
+      defaultValue: this.props.value || 0,
```
amanda-bosson: Nb: If you're setting the default value of `value` to 0 in `defaultProps`, do you need the `||`?

table/src/GroupRow.jsx
```diff
@@ -46,14 +46,18 @@ const GroupRow = ({
...
-      ? (<label
-        className={cx(styles.controlCheckbox, { [styles.disabled]: !groupData.data.length })}
-        onClick={onChange}
-      >
-        {checkboxSvg}
-      </label>)
-      : null);
+    const checkbox = showCheckboxes
+      ? (
+        /* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
```
ben-burlingham: Interesting. Not a huge fan of this global rule but I understand why it's there...interesting to see another valid use case that goes against it.

table/src/HeadRow.jsx
```diff
@@ -11,7 +11,7 @@ export default class HeadRow extends React.Component {
     columnDefinitions: PropTypes.arrayOf(PropTypes.object).isRequired,
     dimensions: PropTypes.shape().isRequired,
     enableDraggableRows: PropTypes.bool.isRequired,
-    isCheckedAll: PropTypes.bool,
+    isCheckedAll: PropTypes.bool.isRequired,
```
ben-burlingham: Are you sure this won't mess up tables w/o checkboxes

table/src/ReactRange.jsx
nickwmurphy: Good point. I'll check that out.

table/src/HeadRow.jsx
```diff
@@ -64,9 +64,14 @@ export default class HeadRow extends React.Component {
       const checkbox = (
         showCheckboxes
           ? (
-            <label className={styles.controlCheckbox} onClick={() => { onCheckAll(!isCheckedAll); }}>
+            /* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
+            <label
+              className={styles.controlCheckbox}
+              onClick={() => { onCheckAll(!isCheckedAll); }}
```
ben-burlingham: Declaring the function inline is a perf hit, since each time the table renders it has to declare the function again.

table/src/Table.jsx
```diff
@@ -385,16 +390,21 @@ export default class extends React.Component {
   columnDefinitionsValidation = (columnDefinitions) => {
     // Add default prop autoResizable: true to column definitions if undefined
```
ben-burlingham: We might consider using this pattern to add more nonexistent props...not sure how it would look, but it feels like something that would come up frequently.  Or, it has come up and it's done somewhere else in the code 🤔

table/src/ReactRange.jsx
nickwmurphy: So I added it because the prop is coming through as NaN at first, but quickly updates to an actual numeric value. This means it does not fall back to its default but still needs 0 as an alternative while mounting/rendering.

table/src/HeadRow.jsx
nickwmurphy: It has a default in the parent component, so should be good!

table/src/GroupRow.jsx
nickwmurphy: It is something we should consider removing from our linter...

table/src/HeadRow.jsx
nickwmurphy: Agreed...follow up

Issue #724, opened 2018-01-12
Table - Columns don't resize when the table becomes smaller


scottwalstead:
## Steps to reproduce
1) Start with a smaller browser window size and have enough columns to fill the width of the table by a small amount so the columns are not auto resized
2) Expand the browser window to grow the size of the table until the columns each start to grow due to auto resizing to fill the width of the table
3) Now shrink the table back to it's original size and notice that the columns don't shrink to their original size

PR #728, opened 2018-01-13
Make columns resize as the table becomes smaller

scottwalstead:
### Summary
When you reduce the table width, the columns should adjust and once the table is scrollable, the columns should be at their original size (see issue below)

### Related Issues
Closes: https://github.com/procore/particles/issues/724

### Testing
Steps to reproduce are documented in https://github.com/procore/particles/issues/724


nickwmurphy:
Scott, I have also solved this in #725. I think I like your solution better. Let's talk tomorrow if you have time!

scottwalstead:
@nickwmurphy sounds good. I'm working remote from home so we can do a Slack call if you'd like and combine the best parts of both of our PRs to solve this issue.

PR #733, opened 2018-01-17
Table: Column Resize Affordance

nickwmurphy:
### Summary
- Added more obvious affordance for column resizing
- Added grab/grabbing cursors for row reordering

col-resize-affordance


nickwmurphy:
Something that indicates to a user that an action can be taken. Adding a screenshot to make it more obvious!

PR #735, opened 2018-01-17
Table Locked Rows

nickwmurphy:
### Summary
- Added prop `lockedRowsPerGroup` to allow for a group of rows (starting at the top of each group) that cannot be reordered nor can any other row be placed within that locked group of rows.
- Some linting and alphabetizing of props
- Also fixes a bug when pressing `enter` while an `InlineSelect` has no data

### Screenshots
table-locked-rows


nickwmurphy:
Performance-wise, this branch should be completely neutral.

In terms of being more granular/specific, I like the idea, but I think we might be better off waiting on the use case for that. Think of this branch as replicating the "locked" functionality that exists in the `TableConfig`. Use case being that "locked" rows are only ever at the top forming a "locked" area that cannot be entered or moved. This will be really useful for our workflow table that has completed steps that need to form a "locked" area from the top down.

nickwmurphy:
That's a good point @brock-noah. I will definitely avoid that in the future.

PR #738, opened 2018-01-19
Table Inline Select onBlur Bug

nickwmurphy:
### Summary
- Fixed a bug where the inline select would not close on blur for React 15 clients


dancingshell:
version bump + readme?

nickwmurphy:
Just got those done!

table/README.md
```diff
@@ -8,6 +8,9 @@ A table component with checkboxes, sticky head, sticky right panel, and horizont
 ![table](https://user-images.githubusercontent.com/8366399/34239862-2189eda6-e5bf-11e7-8254-b4e2d94ccdf4.gif)
 # Change Log
 
+##### 12.4.1
+- Fixed a bug where the inline select would not close on blur
```
dancingshell: might mention that it was only for react 15 clients

table/src/InlinePortal.jsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+/* eslint-disable import/no-extraneous-dependencies */
 import ReactDOM from 'react-dom';
+/* eslint-enable import/no-extraneous-dependencies */
```
dancingshell: I don't know why it says that, its clearly being used 🤷‍♀️

nickwmurphy: 🤷‍♂️

PR #742, opened 2018-01-23
Table Drag and Drop to Final Row, Style Updates

nickwmurphy:
### Summary
- Rows can be dragged and dropped as the bottom row (previously could not)
- Drag and drop style updates (to be more similar to the style/interaction of the `TableConfig`)

### Screenshots
dnd-final-row


table/src/BodyRow.jsx
```diff
@@ -89,30 +92,17 @@ const BodyRow = ({
...
-          <DragHandle />
-        </span>
-      )
-      : cells.unshift(
-        <span
-          className={styles.emptyDragCell}
-          key={`${rowData[idKey]}-drag-cell`}
-        />
-      );
+    const draggable = rowIndex + 1 > lockedRowsPerGroup;
```
brock-noah: `nb` Might be nice as a prop

table/src/Table.jsx
```diff
@@ -526,9 +537,13 @@ export default class extends React.Component {
           {...{
             collapsedIds,
             data,
+            fromGroupIndex,
```
brock-noah: `nb` Looks like `fromGroupIndex` is passed into a few components but never used beyond the `Table` component

table/src/BodyRow.jsx
nickwmurphy: I see what you mean, but since it is only used in this component, I think this helps avoid prop "bloat."

PR #744, opened 2018-01-24
[Select] Search memoization for SelectWithDefaultSearch

dancingshell:
### Summary

- yay memoized results!
- things go faster!

- also refactored state variables a bit

### Testing

* search should essentially work as it did, just be more performant under the hood


select/src/SelectWithDefaultSearch.jsx
```diff
@@ -117,56 +118,74 @@ export default class SelectWithDefaultSearch extends React.Component {
...
+    this.setState({ filteredData: searchResults, searchValue: value });
+    this.props.onSearch({ value, scope });
+  };
+
+  getMemoizedFilterResults = (searchValue) => {
     const { valueField } = this.props;
-    const { groupedData } = this.state;
 
+    if (this.cachedSearchResults.hasOwnProperty(searchValue)) {
+      return this.cachedSearchResults[searchValue];
```
piercefrawley: any reason to use `this.cachedSearchResults` vs `this.state.cachedSearchResults`?

dancingshell: Setting the state triggers a re-render, I think it's lighter weight to store the cache not in state. All we need in state is the current results. But actually this makes me realize that we don't need to store filteredResults in state either... brb going to fix

PR #749, opened 2018-01-25
Overflow Menu Dropdown!

dancingshell:
### Summary

New overflow menu component, using portals and glue components

> Pretend the calculator is an ellipsis. There is currently a bug with the ellipsis icon so I couldn't use it for development, but it is ellipsis in the code
screen shot 2018-01-25 at 1 15 09 pm


overflow-menu/src/MenuItem.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const MenuItem = ({ text, onClick, styles }) => (
+  <div
+    title={text}
```
mboperator: Is there a functional purpose to this text prop? Asking because it seems like we can just use `children` here.

overflow-menu/src/MenuItem.jsx
```diff
@@ -1,26 +1,26 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const MenuItem = ({ text, onClick, styles }) => (
+const MenuItem = ({ title, onClick, styles, className, children }) => (
   <div
-    title={text}
-    className={styles.item}
+    title={title || (typeof children === 'string' ? children : '')}
+    className={`${styles.item} ${className}`}
```
RyanFuentes: I like this pattern of applying the classname of particles to the parent dom element. More particles should apply it.

PR #750, opened 2018-01-25
Tooltip Reposition if New String Content

nickwmurphy:
### Summary
- If content prop is a string, and a different content prop is passed, tooltip will reposition

### Screenshots
Fixing this problem:
tooltip-positioning


tooltip/src/Tooltip.jsx
```diff
@@ -91,7 +91,12 @@ export default class Tooltip extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    if (this.state.show && !prevState.show) {
+    const isContentString = typeof prevProps.content === 'string'
```
dancingshell: nb: could check show prop first, if not show return early and don't do other computations

```js
if (!this.props.show) {
  return;
}

//do rest
```

nickwmurphy: Good call. Will do!

PR #757, opened 2018-01-30
<InlineText />Removed Tab Index so it focuses the input when I click on the pencil (Safari and FireFox)

yoyis3000:
### Summary
On Safari and Firefox, clicking on the pencil icon would just cause the pencil to disappear and it wont focus the input for InlineText. Removing the tab index allowed the container to be focused when the pencil icon is clicked. 

#### Before
inline text before

#### After
inline text after

### Testing
Click on Pencil icon or container and make sure it gets focused.


table/src/InlineText.jsx
```diff
@@ -51,6 +51,8 @@ export default class InlineText extends React.Component {
   onBlur = () => {
     this.props.onBlur();
     this.onSubmit();
+    delete this.container.removeAttribute('tabIndex');
+    console.log('look at me blur');
```
dancingshell: remove log

table/src/InlineText.jsx
```diff
@@ -51,6 +51,8 @@ export default class InlineText extends React.Component {
   onBlur = () => {
     this.props.onBlur();
     this.onSubmit();
+    delete this.container.removeAttribute('tabIndex');
```
dancingshell: if you already removed the `tabIndex` from the element, fo you have to remove it again here?

also the `delete` I'm not sure the delete is doing anything here

yoyis3000: I forgot to _delete_ this.

table/src/InlineText.jsx
yoyis3000: will do!

PR #759, opened 2018-01-30
[GLUE] Readme Clarification

mboperator:
### Summary
Added instructions for completing common development tasks in accordance with feedback received from other developers. The goal here is to lower the barrier to entry for making contributions.

- Are these instructions descriptive enough?
- Could these instructions be more concise without sacrificing clarity?


README.md
```diff
@@ -9,6 +9,10 @@
 - Maintain sandboxes for components.
 - Ship it.
 
+## Key Terms
+- `particle` -- A component that is a part of this repository. Usually contains CORE components mixed with state and business logic.
```
chrisarevir: Right now this is not the case (contains CORE components). If we are moving towards it, we should make it true!

chrisarevir: Also, business logic is present mostly in the client, not so much in the component

README.md
```diff
@@ -17,6 +21,31 @@
...
+### How do I see what a Particle looks like in a sandbox environment?
+1. Follow the steps under "How do I get started?"
+2. Run `npm run sandbox`
+3. Navigate to `localhost:8080`
+
+### How do I view and develop a Particle within Procore without publishing it?
+1. Follow the steps under "How do I get started?"
+2. Run `npm link`
+  > This creates a symlink to the current particle in your user folder's `node_modules` directory
+3. Run `npm run dev`
```
scottwalstead: Cool. Thanks for documenting this. I used to just keep manually typing yarn prod but this is much better.

README.md
```diff
@@ -17,6 +21,31 @@
...
+  > This rebuilds your Particle's `index.js` on file change
+4. In another terminal window, navigate to the client you're developing on (ie: `procore/hydra_clients/core`)
+5. Run `npm link @particles/<your-particle-name>`
+  > This creates a symlink from your client's `node_modules` directory to the local `particle` directory
+
+### What do I do when I'm ready to publish a Particle?
+1. Bump the version
+> Semver is of utmost importance. Please be mindful making good versioning choices.
```
scottwalstead: Maybe we include a reference to https://semver.org/ to help those unfamilar with semantic versioning.

README.md
```diff
@@ -17,6 +21,31 @@
...
+2. Update the README
+> See the README Entry Format section below.
+3. Build the Particle (`npm run prod`)
+4. Commit
+> The `npmignore` file in this repository is pretty accurate so you'll want to commit most things. At minimum you'll want to commit the built `index.js`, as well as the `README`. 
+5. Run `npm publish`
```
scottwalstead: A note saying that there are only some people who can publish particles would be good

scottwalstead: might want to mention the CR step and merging before publishing

PR #761, opened 2018-01-30
[Paginator] No Lato

brock-noah:
### Summary
We are already doing things like `font-family: Roboto, Open Sans, Lato, sans-serif`, let's allow the parent to define the font


ben-burlingham:
Glue team had specifically requested this in design review. No opinion myself.

PR #764, opened 2018-01-31
[Overflow-Menu] fixed onClick called twice bug

dancingshell:
### Summary
- Fixes bug with onClick firing twice


overflow-menu/src/DropdownContent.jsx
```diff
@@ -29,6 +29,7 @@ class DropdownContent extends React.Component {
                   { React.cloneElement(child,
                     {
                       ...child.props,
+                      onClick: undefined,
```
mboperator: NB: Not a fan of this, you can avoid it by destructuring child, ie:
```js
            (({ props: { onClick, ...props }, ...child }) => (
                <List.Item onClick={onItemClick(onClick)}>
                  {React.cloneElement(child, {...props, styles})}
                </List.Item>
              )
            ))}
```
This will break if `child` isn't a React element though. Is there ever a case where `child` is falsey? Seems like potentially unnecessary defensive programming

dancingshell: there was the case when child was falsey and it was causing a bug, as fixed here https://github.com/procore/particles/pull/756

dancingshell:
@mboperator, couldn't come up with a better solution :/

